PR #372, opened 2022-02-14
fix: use sha256 hash as federated identity credential name

### LINKED ISSUE #323: follow up on the char limitation for federated identity credential name.
Service account name can be at most 253 chars and namespace can be at most 63 chars. All this compounded can be long in the max scenario.

Not a blocker for this PR but I think we should follow up on the char limitation for federated identity credential name.

_Originally posted by @aramase in https://github.com/Azure/azure-workload-identity/pull/317#discussion_r784260314_

chewong:
**Reason for Change**:

Fixes #323. 

The name of the federated identity credential is limited 120 characters according to the Azure Portal. Use uuid in the case of 
fic name exceeding 120 characters.

**Requirements**

- [x] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:


pkg/cmd/serviceaccount/create.go
```diff
@@ -50,7 +51,7 @@ func newCreateCmd(authProvider auth.Provider) *cobra.Command {
 	createRunner.AppendPhases(
 		phases.NewAADApplicationPhase(),
 		phases.NewServiceAccountPhase(),
-		phases.NewFederatedIdentityPhase(),
+		phases.NewFederatedIdentityPhase(uuid.New().String),
```
aramase, line 54: The only drawback with using an `uuid` for every request is we can end up creating a fic for same <namespace, name, issuer> when the command is run multiple times with different uid as name. Should we generated a name deterministically based on `<namespace, name, issuer>` tuple that's expected to be unique for an app?

chewong: That's fair. Updated the PR.

aramase (approved):

PR #402, opened 2022-03-25
fix: update proxy-init iptables rule to prevent forwarding loop

aramase:

**Reason for Change**:

- Updates iptables rules to skip redirecting proxy traffic back to itself
- Updates user in `Dockerfile` for proxy 

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/docker/proxy.Dockerfile
+++ b/docker/proxy.Dockerfile
@@ -24,6 +24,6 @@ FROM --platform=${TARGETPLATFORM:-linux/amd64} gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/proxy .
 # Kubernetes runAsNonRoot requires USER to be numeric
-USER 65532:65532
+USER 1501:1501
 
 ENTRYPOINT [ "/proxy" ]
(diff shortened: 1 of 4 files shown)


docker/proxy.Dockerfile
```diff
@@ -24,6 +24,6 @@ FROM --platform=${TARGETPLATFORM:-linux/amd64} gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/proxy .
 # Kubernetes runAsNonRoot requires USER to be numeric
-USER 65532:65532
+USER 1501:1501
```
chewong, line 27: Can we keep `USER` the same or uid `1501` is a special user that we should use here?

aramase: `65532` is commonly defined in Dockerfile to bypass the security scanners. We want to pick a UID that will most likely not conflict with other users running in the pod, so the redirection works as expected. It doesn't necessarily have to be `1501`. I just picked a random one to avoid conflicts with other sidecar proxies (service meshes). Lmk what you think!

chewong (approved):

docker/proxy.Dockerfile
chewong: Makes sense

PR #415, opened 2022-04-06
chore: upgrade to debian-iptables:bullseye-v1.3.0

aramase:

**Reason for Change**:

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:


chewong (approved):

aramase:
It looks like my dev sub hit quota limit with some keyvault hsm tests I was running. That's causing the proxy test (and _probably_ token exchange tests to fail). 

I've created the required artifacts in the CI sub and I'm validating those with a new variable group in https://github.com/Azure/azure-workload-identity/pull/416. Once that's green we can replace the old variable group with new values and retrigger this PR.

Issue #391, opened 2022-03-10
azwi subcommand for pod-identity detection


aramase:
Subcommand to do the following:
1. Detect `AzureIdentity` and `AzureIdentityBinding` in the given namespace.
2. Get the mapping between the pods with `AzureIdentityBinding` in the namespace.
3. Generate the expected deployment and service account yaml for use with workload identity.
    - The outputted yaml can be reviewed and applied by running `kubectl apply`

aramase:
cc @bcho

PR #432, opened 2022-05-03
feat: add `azwi podidentity` detect subcommand

aramase:
**Reason for Change**:

- Adds `azwi podidentity detect` subcommand to detect existing pod identity configuration in a namespace and generate
  1. YAML for resource with proxy sidecar and proxy-init init container
  2. YAML for service account with client-id annotated based on clientID in `AzureIdentity`
    - If the resource is using default service account, we generate a new service account yaml spec

**Requirements**

- [x] squashed commits
- [ ] included documentation

**Issue Fixed**:

fixes https://github.com/Azure/azure-workload-identity/issues/391

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/cmd/podidentity/detect.go
+++ b/pkg/cmd/podidentity/detect.go
@@ -1,13 +1,25 @@
 package podidentity
 
 import (
+	"context"
+	"fmt"
+
+	"github.com/Azure/azure-workload-identity/pkg/kuberneteshelper"
+
+	aadpodv1 "github.com/Azure/aad-pod-identity/pkg/apis/aadpodidentity/v1"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type detectCmd struct {
-	namespace string
-	outputDir string
+	namespace  string
+	outputDir  string
+	kubeClient client.Client
 }
 
 func newDetectCmd() *cobra.Command {
@@ -18,7 +30,7 @@ func newDetectCmd() *cobra.Command {
 		Short: "Detect the existing aad-pod-identity configuration",
 		Long:  "This command will detect the existing aad-pod-identity configuration and generate a sample configuration file for migration to workload identity",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return detectCmd.validate()
+			return detectCmd.prerun()
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return detectCmd.run()
@@ -34,8 +46,10 @@ func newDetectCmd() *cobra.Command {
 	return cmd
 }
 
-func (dc *detectCmd) validate() error {
-	return nil
+func (dc *detectCmd) prerun() error {
+	var err error
+	dc.kubeClient, err = kuberneteshelper.GetKubeClient()
+	return err
 }
 
 func (dc *detectCmd) run() error {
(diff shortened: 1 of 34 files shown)


chewong (commented):
Did a quick review. 

For e2e flow, I am thinking of using bats-core (I can bootstrap it in a separate PR). Right now, we are using github actions to test `azwi serviceaccount create` flow against a kind cluster and validating the labels/annotations using `jq`. With bats-core, we can have more structured test cases to test `azwi`.


What do you think

pkg/cmd/podidentity/detect.go
```diff
@@ -0,0 +1,389 @@
...
+			return err
+		}
+		for i := range pods {
+			// for pods created by higher level constructors like deployment, statefulset, cronjob, job, daemonset, replicaset, replicationcontroller
+			// we can get the owner reference with pod.OwnerReferences
+			ownerFound := false
+			if len(pods[i].OwnerReferences) > 0 {
+				for _, ownerReference := range pods[i].OwnerReferences {
+					// only get the owner reference that was set by the parent controller
+					if ownerReference.Controller != nil && *ownerReference.Controller {
```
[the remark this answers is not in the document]
aramase: Not necessarily. For the core types, I _think_ that would be the case but for custom controllers that add themselves as owners, they **can** set the controller to be non nil and false.

pkg/kuberneteshelper/client.go
```diff
@@ -0,0 +1,59 @@
...
+	_ = clientgoscheme.AddToScheme(scheme)
+
+	// Add aadpodidentity v1 to the scheme.
+	aadPodIdentityGroupVersion := schema.GroupVersion{Group: aadpodv1.GroupName, Version: "v1"}
+	scheme.AddKnownTypes(aadPodIdentityGroupVersion,
+		&aadpodv1.AzureIdentity{},
+		&aadpodv1.AzureIdentityList{},
+		&aadpodv1.AzureIdentityBinding{},
+		&aadpodv1.AzureIdentityBindingList{},
+		&aadpodv1.AzurePodIdentityException{},
```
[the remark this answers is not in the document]
aramase: Correct! With this command, we'll generate config based on `AzureIdentity` and `AzureIdentityBinding`. No changes based on `AzurePodIdentityException`. 

Let me know if you want me to remove `AzurePodIdentityException` from the scheme

pkg/cmd/podidentity/detect.go
```diff
@@ -0,0 +1,389 @@
...
+		sa, err := dc.createServiceAccountFile(localObject.GetServiceAccountName(), localObject.GetName(), clientID)
+		if err != nil {
+			return err
+		}
+		if err = dc.createResourceFile(localObject, sa); err != nil {
+			return err
+		}
+		log.Debugf("generated config for %s, clientID: %s", localObject.GetName(), clientID)
+	}
+
```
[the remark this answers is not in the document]
chewong: LGTM

pkg/cmd/podidentity/detect.go
```diff
@@ -0,0 +1,410 @@
...
+func (dc *detectCmd) createResourceFile(localObject k8s.LocalObject, sa *corev1.ServiceAccount) error {
+	// add the init container to the container list
+	localObject.SetInitContainers(dc.addProxyInitContainer(localObject.GetInitContainers()))
+	// add the proxy container to the container list
+	localObject.SetContainers(dc.addProxyContainer(localObject.GetContainers()))
+	// set the service account name for the object
+	localObject.SetServiceAccountName(sa.GetName())
+	// reset the managed fields to reduce clutter in the output yaml
+	localObject.SetManagedFields(nil)
+	// reset the resource version, uid and other metdata to make the yaml file applyable
```
chewong, line 276:
```suggestion
	// reset the resource version, uid and other metadata to make the yaml file applyable
```

pkg/cmd/podidentity/detect.go
```diff
@@ -0,0 +1,410 @@
...
+		return err
+	}
+
+	labelsToAzureIdentityMap := make(map[string]aadpodv1.AzureIdentity)
+	for _, azureIdentityBinding := range azureIdentityBindings {
+		if azureIdentityBinding.Spec.Selector == "" || azureIdentityBinding.Spec.AzureIdentity == "" {
+			continue
+		}
+		if azureIdentity, ok := azureIdentities[azureIdentityBinding.Spec.AzureIdentity]; ok {
+			labelsToAzureIdentityMap[azureIdentityBinding.Spec.Selector] = azureIdentity
```
chewong, line 142: How are we handling two different AzureIdentityBindings with the same selector?

pkg/cmd/podidentity/detect.go
```diff
@@ -0,0 +1,410 @@
...
+
+	saLabels := make(map[string]string)
+	if sa.GetLabels() != nil {
+		saLabels = sa.GetLabels()
+	}
+	saLabels[webhook.UseWorkloadIdentityLabel] = "true"
+	sa.SetLabels(saLabels)
+
+	// set the annotations for the service account
+	saAnnotations := make(map[string]string)
```
chewong, line 241: Should we also account `azure.workload.identity/service-account-token-expiration` and `azure.workload.identity/tenant-id` in case the user wants to customize them within the same namespace? I am not sure if we can customize `azure.workload.identity/skip-containers` since workloads might not have the same set of skip containers.

chewong (commented):
A few more comments.

pkg/kuberneteshelper/client.go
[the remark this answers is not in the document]
chewong: Let's remove it since we are not using it at the moment

pkg/cmd/podidentity/detect.go
aramase: Great suggestion! Let me add flags for expiration and tenant-id.

I agree we can't provide a good experience for customizing skip containers because that's different across the workloads in the namespace. I was worried about customizing tenant-id too but there is no [cross-tenant option](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/managed-identities-faq#can-i-use-a-managed-identity-to-access-a-resource-in-a-different-directorytenant), so all the identities will be in the same namespace and most likely be in the same tenant as cluster.

pkg/cmd/podidentity/detect.go
aramase: Good question! I've updated `ListAzureIdentityBinding` to first sort the list of `AzureIdentityBinding` and then add to map if not exists. We'll annotate the service account with the clientID from the first `AzureIdentityBinding` -> `AzureIdentity` -> clientID (this behavior is deterministic with the sort) as we can only annotate with 1 clientID.

If the workload is using multiple pod identities with AAD Pod Identity, they will follow the similar model with workload identity. The `AZURE_CLIENT_ID` env var will be just one of the client ids. The workload will most likely not use that env var and instead rely on runtime variables they have configured for which managed identity to use to access which azure resource.

Hope this clarifies your question!

pkg/cmd/podidentity/detect.go
aramase: Done!

pkg/cmd/podidentity/detect.go
aramase: Added a comment with the same explanation!

chewong: `azureIdentityBindings` is a map so iterating it is still indeterministic right? We need to store `azureIdentityBindings` in a slice to preserve the ordering.

aramase: good catch! updated it in https://github.com/Azure/azure-workload-identity/pull/432/commits/5f129a09753fd17f66eb7b5e771c86d806ffd77d and added some unit tests

chewong (approved):

aramase:
cc @bcho

Issue #436, opened 2022-05-04
ModuleNotFoundError: No module named 'azure'


epomatti:
**Describe the bug**

When running the quick-start Python example in the Docker container the following error occurs:

```
ModuleNotFoundError: No module named 'azure'
```

**Steps To Reproduce**

Any of these will have the same issue:

```sh
### Option 1 - Locally from source code:
cd examples/msal-python
docker build . --tag=msal-python
docker run --rm msal-python

### Option 2 - Pulling the image from the registry:
docker run ghcr.io/azure/azure-workload-identity/msal-python

### Option 3 - Kubernetes
kubectl logs quick-start
```

**Expected behavior**

All dependency modules should be available in the Python environment during execution.

**Logs**

```
$ kubectl logs quick-start
Traceback (most recent call last):
  File "/app/main.py", line 4, in <module>
    from azure.keyvault.secrets import SecretClient
ModuleNotFoundError: No module named 'azure'
```

**Environment**

- Kubernetes version (use `kubectl version`): Client: v1.22.5, Server: v1.22.6
- Cloud provider or hardware configuration: Azure
- OS (e.g: `cat /etc/os-release`): WSL Ubuntu 22.04

epomatti:
Merged #437

PR #437, opened 2022-05-04
fix: use debian11 in Dockerfile build image

epomatti:
**Reason for Change**:

Updating the image to use Debian 11.

When running the quick-start Python example the following error occurs.

```
$ kubectl logs quick-start
Traceback (most recent call last):
  File "/app/main.py", line 4, in <module>
    from azure.keyvault.secrets import SecretClient
ModuleNotFoundError: No module named 'azure'
```

The build base image is not the same as the runtime one. Build is buster 10 and runtime is bullseye 11.

Debian has already been bumped in the [original reference code](https://github.com/GoogleContainerTools/distroless/commit/d4d6e705d769eb21bb2a020bb096d0bc624e1d57), and that incidentally fixes the issue as well.

**Requirements**

- [x] squashed commits
- [x] included documentation

**Issue Fixed**: #436

- [x] yes - The existing code already has a reference to the original author.
- [ ] no

### DIFF
--- a/examples/msal-python/Dockerfile
+++ b/examples/msal-python/Dockerfile
@@ -1,10 +1,10 @@
 # ref: https://github.com/GoogleContainerTools/distroless/blob/main/examples/python3-requirements/Dockerfile
 
-FROM debian:buster-slim AS build
+FROM debian:11-slim AS build
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


chewong (approved):
Thank you!

PR #479, opened 2022-06-28
feat: allow setting `azure.workload.identity/use` in annotations

### LINKED ISSUE #473: Support using ServiceAccounts without label
**Is your feature request related to a problem? Please describe.**

At the moment, AAD Workload Identity requires a `azure.workload.identity/use=true` label on Service Accounts used to attach the identity. However, most Helm charts, for example: [external-dns](https://artifacthub.io/packages/helm/external-dns/external-dns) and [ArgoCD](https://artifacthub.io/packages/helm/argo/argo-cd) only provide a variable to set _annotations_ on the Service Account.

Since forking & modifying lots of Helm charts seems cumbersome, this creates major inconveniece when using third-party charts. This also blocks our migration from AAD Pod Identity, which uses Pod labels — a way more common variable in community charts.

**Describe the solution you'd like**

An option to filter Service Accounts based on their annotations instead of relying on a label present would solve the issue. For instance, [AWS IRSA implementation](https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html) only uses annotation on a ServiceAccount.

**Describe alternatives you've considered**

At the moment, we manage ServiceAccounts separately from the charts using them.

aramase:

**Reason for Change**:

-  Allow setting `azure.workload.identity/use` in annotations in addition to labels

**Requirements**

- [x] squashed commits
- [ ] included documentation

**Issue Fixed**:

fixes #473 

- [ ] yes
- [ ] no

**Notes for Reviewers**:
- I'm thinking of updating the docs after `v0.11.0` release

### DIFF
--- a/test/e2e/webhook.go
+++ b/test/e2e/webhook.go
@@ -84,6 +84,12 @@ var _ = ginkgo.Describe("Webhook", func() {
 		validateMutatedPod(f, pod, nil)
 	})
 
+	ginkgo.It("should mutate a deployment pod with a annotated service account", func() {
+		serviceAccount := createServiceAccount(f.ClientSet, f.Namespace.Name, f.Namespace.Name+"-sa", nil, map[string]string{webhook.UseWorkloadIdentityLabel: "true"})
+		pod := createPodUsingDeploymentWithServiceAccount(f, serviceAccount)
+		validateMutatedPod(f, pod, nil)
+	})
+
 	ginkgo.It(fmt.Sprintf("should not mutate selected containers if the pod has %s annotated", webhook.SkipContainersAnnotation), func() {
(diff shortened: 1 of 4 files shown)


test/e2e/webhook.go
```diff
@@ -84,6 +84,12 @@ var _ = ginkgo.Describe("Webhook", func() {
 		validateMutatedPod(f, pod, nil)
 	})
 
+	ginkgo.It("should mutate a deployment pod with a annotated service account", func() {
```
chewong, line 87: nit

```suggestion
	ginkgo.It("should mutate a deployment pod with an annotated service account", func() {
```

chewong (approved):

PR #490, opened 2022-07-05
feat: add proxy --probe and enable lifecycle postStart hook

### LINKED ISSUE #486: Add postStart handler for proxy
This is the outcome of some testing with release charts in https://github.com/Azure/azure-workload-identity/pull/483

If the client makes a connection to IMDS (169.254.169.254) before the proxy is run, then request will fail. The request will continue to fail as long as the client reuses the connection. To prevent this behavior and to make the token request-response more deterministic, we need to ensure the proxy is run first and then the workload. This can be achieved by adding [postStart](https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#define-poststart-and-prestop-handlers) handler to the proxy sidecar container.

1. Implement a healthz API in the proxy
2. Configure `httpGet` post start handler check for proxy

Only after the postStart handler for proxy is executed successfully, the remaining containers will be started.

aramase:

**Reason for Change**:

- Adds a new `--probe` option to the proxy binary
   - When proxy is run with `--probe`, it'll run a readyz probe to check if the proxy is running and ready to serve requests.

**Requirements**

- [x] squashed commits
- [x] included documentation

**Issue Fixed**:

fixes #486 

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/cmd/proxy/main.go
+++ b/cmd/proxy/main.go
@@ -12,6 +12,7 @@ import (
 
 var (
 	proxyPort int
+	probe     bool
 )
 
 func init() {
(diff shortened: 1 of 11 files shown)


pkg/proxy/probe.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Probe checks if the proxy is ready to serve requests.
+func Probe(port int) error {
+	url := fmt.Sprintf("http://%s:%d%s", localhost, port, readyzPathPrefix)
+	return probe(url)
+}
+
+func probe(url string) error {
+	client := &http.Client{
+		Timeout: time.Second * 5,
```
chewong, line 19: ditto regarding constant

cmd/proxy/main.go
```diff
@@ -23,8 +24,21 @@ func main() {
...
+	flag.BoolVar(&probe, "probe", false, "Run a readyz probe on the proxy")
 	flag.Parse()
 
+	// when proxy is run with --probe, it will run a readyz probe on the proxy
+	// this is used in the postStart lifecycle hook to verify the proxy is ready
+	// to serve requests
+	if probe {
+		setupLog := logger.Get().WithName("probe")
+		if err := proxy.Probe(proxyPort); err != nil {
+			setupLog.Error(err, "Failed to probe")
```
chewong, line 36:
```suggestion
			setupLog.Error(err, "failed to probe")
```

chewong (approved):

PR #511, opened 2022-07-31
refactor: Improve Reactor API usage in Java example

ianroberts:
**Reason for Change**:

The current msal-java example exhibits very poor practice in the use of reactive streams.  When a method like `TokenCredential.getToken` returns a `Mono` the caller should expect that the calling thread will not block, and that any potentially long running operations will occur when the resulting `Mono` is _subscribed_, rather than blocking the caller.  Currently `getToken` includes two blocking operations, first the read of the token file from disk and second the call to `CompletableFuture.join()` on the future returned by `ConfidentialClientApplication.acquireToken`.

This PR moves the file read into the `CustomTokenCredential` constructor (which is not a reactive call), and makes `getToken` properly non-blocking.

**Requirements**

- [x] squashed commits
- [ ] included documentation (n/a)
- [ ] added unit tests and e2e tests (if applicable). (n/a)

Does this change contain code from or inspired by another project?

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/examples/msal-java/pom.xml
+++ b/examples/msal-java/pom.xml
@@ -13,8 +13,8 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <maven.compiler.source>1.7</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
   </properties>
 
   <dependencies>
@@ -37,6 +37,12 @@
       <groupId>com.microsoft.azure</groupId>
       <artifactId>msal4j</artifactId>
       <version>1.11.2</version>
+    </dependency>
+    <!-- in a real application, you would use a proper logging implementation such as log4j2 or logback -->
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-simple</artifactId>
+      <version>1.7.32</version>
     </dependency>
   </dependencies>
   <build>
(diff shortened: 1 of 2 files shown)


ianroberts:
(Unsure whether this is `docs:` or `fix:` - it's a fix to sample code that is part of the documentation rather than the actual implementation.)

examples/msal-java/pom.xml
```diff
@@ -13,8 +13,8 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <maven.compiler.source>1.7</maven.compiler.source>
-    <maven.compiler.target>1.7</maven.compiler.target>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
```
ianroberts, line 17: Changed minimum language level to 1.8 in order to allow a lambda expression in `Mono.defer`

ianroberts: I don't think that's unreasonable given msal4j itself requires at least 1.8

aramase (commented):
@ianroberts Thank you for the PR!

I tried out the changes and it works as expected. However, I do see log messages about failing to load the `org.slf4j.impl.StaticLoggerBinder` class. What is the best way to silence those logs? 

```
➜ kl quick-start
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
successfully got secret, secret=Hello!
```

ianroberts:
The way to silence those warnings is to add a dependency on a suitable logging implementation - the simplest option for example code like this is slf4j-simple, I've added that to the pom and re-based.

ianroberts:
If the logging is still too noisy then I'll change it to depend on logback and include an "error messages only" logging configuration file.

aramase (approved):

aramase (commented):
The current changes look good! Thank you for the PR 🎉

PR #526, opened 2022-08-12
docs: update docs to use azure cli for federated identity credential

salaxander:
Update the docs to use the `az` cli for the federated identity credential section rather than `az rest`


docs/book/src/topics/federated-identity-credential.md
```diff
@@ -44,11 +44,10 @@ azwi serviceaccount delete phase federated-identity \
 To create a federated identity credential, login to [Azure Cloud Shell][1] and run the following commands:
 
 ```bash
-# Get the client and object ID of the AAD application
-export APPLICATION_CLIENT_ID="$(az ad sp list --display-name "${APPLICATION_NAME}" --query '[0].appId' -otsv)"
-export APPLICATION_OBJECT_ID="$(az ad app show --id "${APPLICATION_CLIENT_ID}" --query id -otsv)"
+# Get the object ID of the AAD application
```
aramase, line 47: In the docs we show how to set up a federated identity credential using:
1. `az rest` in cloud shell
2. `azwi` CLI
3. Azure portal UI

Could we add a 4th option with just azure cli and add these in that section? That way the user can choose what option works best for them. 

nit: do we need cloud shell for these commands?

sozercan: cloud shell is updated with latest az cli (.39 now) so we shouldn't need `az rest` anymore i think

salaxander: Yeah I would think we would want to drop the `az rest` method at this point right? I'm guessing cloud shell was suggested because it provides an up-to-date environment. I can add a note there to use cloud shell or a local terminal

salaxander:
@aramase How does this iteration look to you?

docs/book/src/topics/federated-identity-credential.md
```diff
@@ -41,14 +41,13 @@ azwi serviceaccount delete phase federated-identity \
...
 
-To create a federated identity credential, login to [Azure Cloud Shell][1] and run the following commands:
+A federated identity credential can also be created using the `az` CLI. This can either be done in a local terminal session, or using [Azure Cloud Shell][1]. Use the `az` CLI to run the following commands:
 
 ```bash
-# Get the client and object ID of the AAD application
-export APPLICATION_CLIENT_ID="$(az ad sp list --display-name "${APPLICATION_NAME}" --query '[0].appId' -otsv)"
-export APPLICATION_OBJECT_ID="$(az ad app show --id "${APPLICATION_CLIENT_ID}" --query id -otsv)"
+# Get the object ID of the AAD application
+export APPLICATION_OBJECT_ID="az ad app list --display-name "${APPLICATION_NAME}" --query '[0].id' -otsv"
```
aramase, line 48:
```suggestion
export APPLICATION_OBJECT_ID="$(az ad app list --display-name "${APPLICATION_NAME}" --query '[0].id' -otsv)"
```

docs/book/src/topics/federated-identity-credential.md
```diff
@@ -60,19 +59,19 @@ cat <<EOF > body.json
...
 
-az rest --method POST --uri "https://graph.microsoft.com/beta/applications/${APPLICATION_OBJECT_ID}/federatedIdentityCredentials" --body @body.json
+az ad app federated-credential create --id $APPLICATION_OBJECT_ID --parameters params.json
 ```
 
 To delete a federated identity credential, the federated identity credential ID needs to be obtained with the following command:
 
 ```bash
-az rest --method GET --uri "https://graph.microsoft.com/beta/applications/${APPLICATION_OBJECT_ID}/federatedIdentityCredentials"
+export FIC_ID="$(az ad app federated-credential list --id "${APPLICATION_OBJECT_ID}")"
```
aramase, line 68:
```suggestion
az ad app federated-credential list --id "${APPLICATION_OBJECT_ID}"
```

list returns an array of federated credential, so we should let the users pick the one they want to delete

aramase (changes requested):
Almost there!

aramase (approved):

PR #531, opened 2022-08-15
chore: update to go 1.19

aramase:

**Reason for Change**:

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:


.github/workflows/azwi-build.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-go@v3
         with:
-          go-version: "^1.18"
+          go-version: "^1.19"
```
sozercan, line 30: I believe caret will upgrade to 1.19, would we want to pin these like `1.19`?

aramase: Updated it to `1.19`

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/azure-workload-identity
 
-go 1.18
+go 1.19
```
sozercan, line 3: why are we bumping go directives? do we require 1.19 to build?

examples/msal-go/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/azure-workload-identity/example/msal-go
 
-go 1.18
+go 1.19
```
sozercan, line 3: same here

sozercan (approved):
added a comment, otherwise LGTM

go.mod
aramase: we're bumping the base image used to build to 1.19, so this change would be consistent with that!

sozercan: Docs are pretty ambiguous about this but it's meant as a min version so if we were using a specific feature in 1.19 then we would update this, otherwise keep the same version.

PR #544, opened 2022-08-30
feat: introduce security context for deployment

Jaxwood:
Signed-off-by: Jacob Lorenzen <jacob@lorenzen.me>

**Reason for Change**:
Introduce Pod security context to helm chart

Yes

Yes

**Requirements**

- [x] squashed commits
- [x] included documentation

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:
I noticed container security context is set through code - so I haven't introduced the possibility to override it via helm as I assume it has to be configured in a certain way..


aramase (changes requested):
`securityContext` is already defined for the webhook pods https://github.com/Azure/azure-workload-identity/blob/46c5137ef22d824440a760c707edd24115f46153/manifest_staging/charts/workload-identity-webhook/templates/azure-wi-webhook-controller-manager-deployment.yaml#L69-L79

Jaxwood:
> `securityContext` is already defined for the webhook pods
> 
> https://github.com/Azure/azure-workload-identity/blob/46c5137ef22d824440a760c707edd24115f46153/manifest_staging/charts/workload-identity-webhook/templates/azure-wi-webhook-controller-manager-deployment.yaml#L69-L79

@aramase is it both on container and pod level?

PR #561, opened 2022-09-13
chore: Load keyvault url from environment variable in example

karataliu:
**Reason for Change**:
Make it easy to test in sovereign clouds

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/examples/msal-go/main.go
+++ b/examples/msal-go/main.go
@@ -12,10 +12,12 @@ import (
 )
 
 func main() {
-	keyvaultName := os.Getenv("KEYVAULT_NAME")
-	secretName := os.Getenv("SECRET_NAME")
-
-	keyvaultURL := fmt.Sprintf("https://%s.vault.azure.net/", keyvaultName)
+	keyvaultURL := os.Getenv("KEYVAULT_URL")
+        if keyvaultURL == "" {
+                keyvaultName := os.Getenv("KEYVAULT_NAME")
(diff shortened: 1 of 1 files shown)


aramase (approved):

examples/msal-go/main.go
```diff
@@ -12,10 +12,13 @@ import (
 )
 
 func main() {
-	keyvaultName := os.Getenv("KEYVAULT_NAME")
-	secretName := os.Getenv("SECRET_NAME")
-
-	keyvaultURL := fmt.Sprintf("https://%s.vault.azure.net/", keyvaultName)
+	keyvaultURL := os.Getenv("KEYVAULT_URL")
+        if keyvaultURL == "" {
+                keyvaultName := os.Getenv("KEYVAULT_NAME")
```
aramase, line 17: Could you move `keyvaultName` outside the if loop?

aramase (changes requested):

examples/msal-go/main.go
karataliu: build fixed. ptal

aramase: > I suppose we can remove keyvaultName in future after the doc is changed to use keyvaultURL?

For this change, we should update the env var used across all the language examples here: https://github.com/Azure/azure-workload-identity/tree/main/examples

aramase (approved):

PR #563, opened 2022-09-15
docs: Use KEYVAULT_URL in quick start guide

karataliu:
**Reason for Change**:
Update doc to use KEYVAULT_URL, so as to support sovereign clouds

- [ ] yes
- [x] no

**Notes for Reviewers**:
Follow-up #561

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -238,6 +238,8 @@ spec:
     - image: ghcr.io/azure/azure-workload-identity/msal-go
       name: oidc
       env:
+      - name: KEYVAULT_NAME
+        value: ${KEYVAULT_NAME}
       - name: KEYVAULT_URL
         value: ${KEYVAULT_URL}
       - name: SECRET_NAME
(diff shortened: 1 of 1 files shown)


docs/book/src/quick-start.md
```diff
@@ -237,8 +238,8 @@ spec:
     - image: ghcr.io/azure/azure-workload-identity/msal-go
       name: oidc
       env:
-      - name: KEYVAULT_NAME
-        value: ${KEYVAULT_NAME}
+      - name: KEYVAULT_URL
+        value: ${KEYVAULT_URL}
```
aramase, line 242: Let's add `KEYVAULT_URL` as an additional environment variable and not remove `KEYVAULT_NAME`. If old version of `msal-go` is used, it'll not work with `KEYVAULT_URL`. We can remove `KEYVAULT_NAME` after a few releases and after every [example](https://github.com/Azure/azure-workload-identity/tree/main/examples) is updated to use `KEYVAULT_URL`.

aramase (changes requested):

aramase:
@karataliu could you also update the commit to say `docs: Use KEYVAULT_URL in quick start guide` as this a documentation update?

docs/book/src/quick-start.md
karataliu: updated.

aramase (approved):

PR #606, opened 2022-10-21
chore: Remove `creationTimestamp: null` from config

karataliu:
**Reason for Change**:
Simplify deployment

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -3,7 +3,6 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  creationTimestamp: null
   name: manager-role
 rules:
 - apiGroups:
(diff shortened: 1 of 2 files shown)


aramase (changes requested):
Changes to `charts` and `deploy` dir is only done at the time of release. Could you revert that change from the files in those 2 dirs?

karataliu:
updated.

aramase (changes requested):
@karataliu The `creationTimestamp` is added automatically when generating the manifests based on `kubebuilder` markers. I don't think we can remove this manually as it'll get added back when we run `make manifests` as part of every release. Is there a reason this needs to be removed? If not, I think we can leave it as is. LMK!

sozercan:
I beieve this needs to be addressed to remove these https://github.com/kubernetes-sigs/controller-tools/issues/402

karataliu:
@aramase 

I'd like to keep manifest as precise as possible (for easy sync), so it's good to keep out those unnecessary lines.

It's ok to keep it there if it's autogenerated. Please close then.

Issue #145, opened 2021-08-12
Fallback to AZURE_CLIENT_ID env var if no `client_id` query param in token request


aramase:
This is still up for debate so I didn't want to add this as part of this PR. I'll add a follow-up PR once we make a decision on this.

_Originally posted by @aramase in https://github.com/Azure/aad-pod-managed-identity/pull/142#discussion_r688030472_

polivbr:
I realize this issue is a bit old, but are there any further thoughts on this? The current behavior is making it challenging to move certain third-party applications (such as Pulumi) from aad-pod-identity to Workflow Identity.

aramase:
> I realize this issue is a bit old, but are there any further thoughts on this? The current behavior is making it challenging to move certain third-party applications (such as Pulumi) from aad-pod-identity to Workflow Identity.

@polivbr Can you provide more details on why this is blocking your move to workload identity?

polivbr:
In the particular case of Pulumi, the Azure provider currently doesn't handle OIDC auth, so I have to use MSI. When configured to use MSI, rather than looking for an AZURE_CLIENT_ID environment variable, it looks for ARM_CLIENT_ID. If it's not set, it doesn't pass the client_id query param, as it assumes it's not needed, which worked fine with aad-pod-identity but breaks with workload identity. 

This isn't a showstopper as I can work around it, but it cost me many, many hours trying to debug exactly why it was failing.

jinhong-:
Am currently stuck on this. Tried to integrate fluxcd Image update automation but `image-reflector-controller` pod would spit out a message `The client_id parameter is required.` as they do not include `client_id` when making a call to the metadata endpoint

polivbr:
This is now causing me issues with services that use the rust sdk. Its identity library doesn't currently support OIDC and code that uses its default token credentials won't pick up a client_id.

Is there a reason to not enable this fallback? Seems like not doing so causes way more issues than it could possibly be solving. It has literally broken every app I've attempted to deploy and now I'm faced with one for which I don't have a workaround.

aramase:
I'm ok with adding the fallback to `AZURE_CLIENT_ID`. This issue was opened to solicit feedback to see if it's required. I'll include this in the next release.

polivbr:
Excellent. Thank you!

PR #628, opened 2022-11-08
feat: Fallback to `AZURE_CLIENT_ID` env var if no `client_id` query param in token request

aramase:

**Reason for Change**:

If clientID not found in request in the `client_id` parameter, then we default to the `AZURE_CLIENT_ID` env var if present. This is to keep consistent with the current behavior in pod identity v1 where we default the client id to the one in `AzureIdentity` if clientID isn't provided as part of the token request.

**Requirements**

- [x] squashed commits
- [x] included documentation

**Issue Fixed**:

fixes https://github.com/Azure/azure-workload-identity/issues/145

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -107,11 +107,16 @@ func (p *proxy) msiHandler(w http.ResponseWriter, r *http.Request) {
 	p.logger.Info("received token request", "method", r.Method, "uri", r.RequestURI)
 	w.Header().Set("Server", userAgent)
 	clientID, resource := parseTokenRequest(r)
-	// TODO (aramase) should we fallback to the clientID in the annotated service account
-	// if clientID not found in request? This is to keep consistent with the current behavior
-	// in pod identity v1 where we default the client id to the one in AzureIdentity.
+	// if clientID not found in request, then we default to the AZURE_CLIENT_ID if present.
+	// This is to keep consistent with the current behavior in pod identity v1 where we
+	// default the client id to the one in AzureIdentity.
 	if clientID == "" {
-		http.Error(w, "The client_id parameter is required.", http.StatusBadRequest)
+		p.logger.Info("client_id not found in request, defaulting to AZURE_CLIENT_ID", "method", r.Method, "uri", r.RequestURI)
+		clientID = os.Getenv(webhook.AzureClientIDEnvVar)
+	}
+
+	if clientID == "" {
+		http.Error(w, "The client_id parameter or AZURE_CLIENT_ID environment variable must be set", http.StatusBadRequest)
 		return
 	}
 	if resource == "" {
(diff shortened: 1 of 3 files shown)


test/e2e/proxy_test.go
```diff
@@ -72,4 +72,57 @@ var _ = ginkgo.Describe("Proxy [LinuxOnly] [AKSSoakOnly] [Exclude:Arc]", func()
...
+			nil,
+			proxyAnnotations,
+			true,
+		)
+
+		pod, err := createPod(f.ClientSet, pod)
+		framework.ExpectNoError(err, "failed to create pod %s in %s", pod.Name, namespace)
+		defer f.ClientSet.CoreV1().Pods(namespace).Delete(context.TODO(), pod.Name, metav1.DeleteOptions{})
+
+		// output proxy and proxy init logs for debugging
```
enj, line 107: Should this only happen when the test has failed?

test/e2e/proxy_test.go
```diff
@@ -72,4 +72,57 @@ var _ = ginkgo.Describe("Proxy [LinuxOnly] [AKSSoakOnly] [Exclude:Arc]", func()
...
+		for _, container := range []string{busybox1, busybox2} {
+			framework.Logf("validating that %s in %s has acquired a valid AAD token via the proxy using AZURE_CLIENT_ID", container, pod.Name)
+			gomega.Eventually(func() bool {
+				stdout, err := e2epod.GetPodLogs(f.ClientSet, namespace, pod.Name, container)
+				if err != nil {
+					framework.Logf("failed to get logs from container %s in %s/%s: %v. Retrying...", container, namespace, pod.Name, err)
+					return false
+				}
+				framework.Logf("stdout: %s", stdout)
+				return strings.Contains(stdout, `"environmentName": "AzureCloud"`)
```
enj, line 124: Could you add a comment describing what this means?  I do not have enough context to understand what this assertion is checking.

test/e2e/proxy_test.go
aramase: We have been dumping logs for success/failure runs so the logs are available for reference. The webhook logs are uploaded to artifacts at the end of the test run.

pkg/proxy/proxy.go
```diff
@@ -111,6 +111,7 @@ func (p *proxy) msiHandler(w http.ResponseWriter, r *http.Request) {
 	// This is to keep consistent with the current behavior in pod identity v1 where we
 	// default the client id to the one in AzureIdentity.
 	if clientID == "" {
+		p.logger.Info("client_id not found in request, defaulting to AZURE_CLIENT_ID", "method", r.Method, "uri", r.RequestURI)
```
enj, line 114: This line looks like it's in the wrong commit?

pkg/proxy/proxy_test.go
```diff
@@ -46,7 +46,7 @@ func TestProxy_MSIHandler(t *testing.T) {
 			name:               "client_id is missing",
 			path:               `/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fvault.azure.net%2F`,
 			expectedStatusCode: http.StatusBadRequest,
-			expectedBody:       "The client_id parameter is required.\n",
+			expectedBody:       "The client_id parameter or AZURE_CLIENT_ID environment variable must be set\n",
```
enj, line 49: This line looks like it's in the wrong commit?

pkg/proxy/proxy.go
aramase: should be fixed now!

enj (approved):

Issue #624, opened 2022-11-02
AAD token received via the side car is valid for 0 seconds


akshaysngupta:
**Describe the bug**
We are seeing a weird issue where AAD token issued for federated identity (used by a pod with sidecar) are only valid for 0 second. While if using regular azure VM with IMDS endpoint returns a token is valid for 24 hours.

**Steps To Reproduce**
* Use sidecar to intercept IMDS calls (migration scenario).
* Use curl command to send IMDS token requests.

**Expected behavior**
Token should be valid for 24 hours.

**Logs**
Token validity when curling the IMDS endpoint
On pod with Side car
![image](https://user-images.githubusercontent.com/5294363/199375197-93cc6ee2-d194-4e4e-b291-a7890bfc205e.png)

Using Pod identity addon (gives a token valid for longer duration)
![image](https://user-images.githubusercontent.com/5294363/199543343-35b90048-69ec-4944-a1f2-97dde9fa910b.png)

On regular azure VM
![image](https://user-images.githubusercontent.com/5294363/199375214-c79fa2bf-7509-4966-8b98-f0709460fbf2.png)

**Environment**
Public Cloud
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`): Linux
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug): CNI
- Others: 

**Additional context**

PR #630, opened 2022-11-08
fix: set `expires_in` the proxy token response

aramase:

**Reason for Change**:

- [msal-go](https://github.com/AzureAD/microsoft-authentication-library-for-go) returns [AuthResult](https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go@v0.7.0/apps/internal/base#AuthResult) which doesn't contain the `expires_in` value. Computing that in the proxy based on current time and expires_on. This change is to make the response consistent with the IMDS response and also AAD response for token request.

This is the AAD response:

```json
{
    "token_type": "Bearer",
    "expires_in": 86399,
    "ext_expires_in": 86399,
    "access_token": "<token>"
}
```
xref: https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#successful-response-2

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

fixes https://github.com/Azure/azure-workload-identity/issues/624

- [ ] yes
- [ ] no

**Notes for Reviewers**:
Opened https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/358 to request adding this as part of AuthResult, so we don't have to compute it.

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -200,11 +200,12 @@ func doTokenRequest(ctx context.Context, clientID, resource, tenantID, authority
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to acquire token")
 	}
-
 	return &token{
 		AccessToken: result.AccessToken,
 		Resource:    resource,
 		Type:        "Bearer",
+		// -10s is to account for current time changes between the calls
+		ExpiresIn: json.Number(strconv.FormatInt(int64(time.Until(result.ExpiresOn)/time.Second)-10, 10)),
 		// There is a difference in parsing between the azure sdks and how azure-cli works
 		// Using the unix time to be consistent with response from IMDS which works with
 		// all the clients.


akshaysngupta (approved):
Thanks! Is there a way to test this with the addon ?

aramase:
> Thanks! Is there a way to test this with the addon ?

Once this is merged, you can use the nightly image build from [here](https://github.com/Azure/azure-workload-identity/pkgs/container/azure-workload-identity%2Fproxy) in your pod spec by manually editing the pod. This will be part of next release (2 weeks from now) post which AKS will update the add-on to use the new release.

pkg/proxy/proxy.go
```diff
@@ -200,11 +206,16 @@ func doTokenRequest(ctx context.Context, clientID, resource, tenantID, authority
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to acquire token")
 	}
+	expiresIn, err := parseExpiresIn(strconv.FormatInt(result.ExpiresOn.UTC().Unix()-time.Now().UTC().Unix(), 10))
```
enj, line 209: This is susceptible to clock skew between the token creator and the proxy's current time.

pkg/proxy/proxy.go
```diff
@@ -200,11 +206,16 @@ func doTokenRequest(ctx context.Context, clientID, resource, tenantID, authority
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse expires in")
+	}
 
 	return &token{
 		AccessToken: result.AccessToken,
 		Resource:    resource,
 		Type:        "Bearer",
+		ExpiresIn:   expiresIn,
```
enj, line 218: Reading the code I see that clock skew is not an issue here because `ExpiresOn` is computed locally.  I think this approach is simpler (I took off 10 seconds to account for the current time changing between calls):

```suggestion
		ExpiresIn:   json.Number(strconv.FormatInt(int64(time.Until(result.ExpiresOn)/time.Second) - 10, 10)),
```

aramase: Updated and removed `parseExpiresIn` as we don't need to handle different format. As per the AAD doc, `expires_in` is in seconds.

pkg/proxy/proxy.go
```diff
@@ -200,11 +200,11 @@ func doTokenRequest(ctx context.Context, clientID, resource, tenantID, authority
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to acquire token")
 	}
-
 	return &token{
 		AccessToken: result.AccessToken,
 		Resource:    resource,
 		Type:        "Bearer",
+		ExpiresIn:   json.Number(strconv.FormatInt(int64(time.Until(result.ExpiresOn)/time.Second)-10, 10)),
```
enj, line 207: Probably worth a comment on why we are doing this at all and the `-10` bit.

enj (approved):

sozercan (approved):

PR #637, opened 2022-11-14
docs: Update link to AKS docs for OIDC Issuer

AaronCrawfis:
**Reason for Change**:
The link to the Azure docs is semi-broken due to the change in the header name. It is also region specific instead of generalized, and uses the old docs subdomain instead of learn.

- Remove preview suffix
- Remove region-specific path
- Switch to new learn platform instead of docs

**Requirements**

- [x] squashed commits
- [x] included documentation

**Issue Fixed**:

- [ ] yes
- [x] no

**Notes for Reviewers**:


aramase (approved):
Thank you for the PR @AaronCrawfis!

Issue #647, opened 2022-11-28
Pod and namespace stuck in terminating state


nilsgstrabo:
**Describe the bug**
Pods with finalizers (e.g. "batch.kubernetes.io/job-tracking" when created by the Job controller in Kubernetes >= 1.23) are stuck in Terminating when deleting the Pod's namespace.

Deleting the namespace will delete the Job, the Pod and the ServiceAccount. Because of the finalizer, the Pod is only flagged as deleted. When the Job controller tries to remove the finalizer from the Pod, the update is rejected by the Azure Workload Identity mutating webhook because the ServiceAccount no longer exists. We have to stop the webhook to allow the Job controller to remove the finalizer. Once the finalizer has been removed, we can start the webhook again.

**Steps To Reproduce**

Create a namespace
`kubectl create ns awi`

Create a ServiceAccount:
```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    azure.workload.identity/client-id: <a valid client id>
  labels:
    azure.workload.identity/use: "true"
  name: my-sa
  namespace: awi
EOF
```

Create a Job:
```
cat <<EOF | kubectl apply -f -
apiVersion: batch/v1
kind: Job
metadata:
  name: quick-start-job
  namespace: awi
spec:
  template:
    metadata:
      labels:
        azure.workload.identity/use: "true"
    spec:
      serviceAccountName: my-sa
      restartPolicy: Never
      containers:
        - image: busybox
          args: [/bin/sh, -c, 'sleep 300']
          name: busybox
EOF
```

Delete the namespace before the sleep 300 expires
`kubectl delete ns awi`

The Pod and Namespace is now stuck in Terminating until the webhook is stopped.

**Expected behavior**

Not sure what the best/correct way to handle situations like this is. Perhaps ignore validation for pods with a deletionTimestamp?

**Logs**

```
E1128 15:09:24.741764       1 webhook.go:115] handler "msg"="failed to get service account" "error"="serviceaccounts \"my-sa\" not found" "namespace"="awi" "pod"="quick-start-job-rnpqt" "service-account"="my-sa"
E1128 15:09:24.795016       1 webhook.go:115] handler "msg"="failed to get service account" "error"="serviceaccounts \"my-sa\" not found" "namespace"="awi" "pod"="quick-start-job-rnpqt" "service-account"="my-sa"
[...]

PR #652, opened 2022-11-29
fix: remove `UPDATE` operation from mutatingwebhookconfiguration

aramase:

**Reason for Change**:

- If the pod is in terminating state, we return early in the handler as mutation/check doesn't apply to those.

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

#647 

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -32,7 +32,7 @@ var (
 	ProxyImageVersion string
 )
 
-// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=ignore,groups="",resources=pods,verbs=create;update,versions=v1,name=mutation.azure-workload-identity.io,sideEffects=None,admissionReviewVersions=v1;v1beta1,matchPolicy=Equivalent
+// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=ignore,groups="",resources=pods,verbs=create,versions=v1,name=mutation.azure-workload-identity.io,sideEffects=None,admissionReviewVersions=v1;v1beta1,matchPolicy=Equivalent
 // +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch
 
 // this is required for the webhook server certs generated and rotated as part of cert-controller rotator
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -24,7 +24,6 @@ webhooks:
     - v1
     operations:
     - CREATE
-    - UPDATE
     resources:
     - pods
   sideEffects: None
--- a/manifest_staging/charts/workload-identity-webhook/templates/azure-wi-webhook-mutating-webhook-configuration-mutatingwebhookconfiguration.yaml
+++ b/manifest_staging/charts/workload-identity-webhook/templates/azure-wi-webhook-mutating-webhook-configuration-mutatingwebhookconfiguration.yaml
@@ -30,7 +30,6 @@ webhooks:
     - v1
     operations:
     - CREATE
-    - UPDATE
     resources:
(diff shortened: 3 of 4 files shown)


pkg/webhook/webhook.go
```diff
@@ -102,6 +102,12 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
 	}
 
 	logger := log.Log.WithName("handler").WithValues("pod", pod.Name, "namespace", pod.Namespace, "service-account", serviceAccountName)
+
+	if !pod.GetDeletionTimestamp().IsZero() {
+		logger.Info("pod is being terminated")
+		return admission.Allowed("pod is being terminated")
+	}
```
enj, line 109: Should we also skip pods that are in the succeeded or failed phase?

pkg/webhook/webhook.go
```diff
@@ -102,6 +102,12 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+
 	// get service account associated with the pod
 	serviceAccount := &corev1.ServiceAccount{}
 	if err = m.client.Get(ctx, types.NamespacedName{Name: serviceAccountName, Namespace: pod.Namespace}, serviceAccount); err != nil {
```
enj, line 113: Should we also make the `m.reader.Get` branch below return `admission.Allowed` when the SA is not found?  Or maybe just keep going with some "default" idea of what the SA state is?

pkg/webhook/webhook.go
```diff
@@ -102,6 +102,12 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
 	}
 
 	logger := log.Log.WithName("handler").WithValues("pod", pod.Name, "namespace", pod.Namespace, "service-account", serviceAccountName)
+
+	if !pod.GetDeletionTimestamp().IsZero() {
```
enj, line 106: Possibly dumb question: why are we invoking the webhook on non-create requests?

enj: i.e. can we drop this line?

https://github.com/Azure/azure-workload-identity/blob/e00e0d315583c827dab98d993dd8243c626e1c97/charts/workload-identity-webhook/templates/azure-wi-webhook-mutating-webhook-configuration-mutatingwebhookconfiguration.yaml#L32

pkg/webhook/webhook.go
aramase: I think we can do this in `v1.0.0-alpha` release when we are sure every pod request that comes for admission review is intended. Now we rely on the service account to know if this pod needs to be mutated.

pkg/webhook/webhook.go
aramase: We can drop the `UPDATE` operation because we don't reset the env variables/ volume if they already exist. Just `CREATE` operation would be sufficient and we don't need a pod deletion check any more. I've updated the PR with that!

pkg/webhook/webhook.go
aramase: opened an issue to track this change

pkg/webhook/webhook_test.go
```diff
@@ -744,6 +744,7 @@ func TestHandle(t *testing.T) {
 	tests := []struct {
 		name               string
 		serviceAccountName string
+		deletionTimeStamp  *metav1.Time
```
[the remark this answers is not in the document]
aramase: thanks, removed!

enj (approved):

ringerc:
This change is described as 

> If the pod is in terminating state, we return early in the handler as mutation/check doesn't apply to those.

but in reality, the change appears to be to not fire the webhook on pod `UPDATE` at all. That's a bit confusing.

PR #698, opened 2023-01-12
fix: explicitly set runAsNonRoot to false (#697)

### LINKED ISSUE #697: proxy-init container doesn't set `runAsNonRoot: false`
**Describe the bug**
When a pod has a `securityContext`, all its containers inherit that and have the option of overriding each setting in that map. If this `podSecurityContext` includes `runAsNonRoot: true`, then the `proxy-init`container fails to run, because it requires running as root. The webhook does set `runAsUser: 0`, so I assume this is an oversight rather than a deliberate choice.

**Steps To Reproduce**
Create a pod configured to use workload identity and the annotation `azure.workload.identity/inject-proxy-sidecar: "true"`, and include a `spec.securityContext` that includes `runAsNonRoot: true`. I ran into this installing the Hashicorp Vault Helm chart. Vault doesn't support MSAL yet, so it requires the proxy to use WI.

**Expected behavior**
The `proxy-init` `initContainer` starts and executes correctly.

**Logs**
Scrolled off-screen, I'm afraid; the events in the output of `kubectl describe pod` are where the complaint can easily be found.

**Environment**

- Kubernetes version (use `kubectl version`): server=1.23.12
- Cloud provider or hardware configuration: Azure AKS

**Additional context**
I have a fix which works for me. I'll submit it as a PR.

dhduvall:
**Reason for Change**:
See #697.

**Requirements**

- [x] squashed commits
- [ ] included documentation

**Issue Fixed**:
Fixes #697 

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -248,8 +248,9 @@ func (m *podMutator) injectProxyInitContainer(containers []corev1.Container, pro
 				Add:  []corev1.Capability{"NET_ADMIN"},
 				Drop: []corev1.Capability{"ALL"},
 			},
-			Privileged: pointer.BoolPtr(true),
-			RunAsUser:  pointer.Int64Ptr(0),
+			Privileged:   pointer.BoolPtr(true),
+			RunAsNonRoot: pointer.BoolPtr(false),
+			RunAsUser:    pointer.Int64Ptr(0),
 		},
 		Env: []corev1.EnvVar{{
 			Name:  ProxyPortEnvVar,
--- a/pkg/webhook/webhook_test.go
+++ b/pkg/webhook/webhook_test.go
@@ -1251,8 +1251,9 @@ func TestInjectProxyInitContainer(t *testing.T) {
 				Add:  []corev1.Capability{"NET_ADMIN"},
 				Drop: []corev1.Capability{"ALL"},
 			},
-			Privileged: pointer.BoolPtr(true),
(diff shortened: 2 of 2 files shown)


aramase (commented):
Thank you for the PR @dhduvall! 

could you update the commit message to be [semantic](https://github.com/Azure/azure-workload-identity/blob/main/.github/semantic.yml) `fix: explicitly set runAsNonRoot to false (#697)`.

dhduvall:
> Thank you for the PR @dhduvall!
> 
>     1. could you update the commit message to be [semantic](https://github.com/Azure/azure-workload-identity/blob/main/.github/semantic.yml) `fix: explicitly set runAsNonRoot to false (#697)`.
> 
>     2. Rebase on `main` branch

Both of those should be done in my latest push.  Thanks for the quick response!

aramase (commented):
Could you also update the struct in the unit tests?

https://github.com/Azure/azure-workload-identity/blob/f41afae1a2b3659f3d244103e6009253f292ab7f/pkg/webhook/webhook_test.go#L1254-L1255

aramase:
@dhduvall Have you agreed to the CLA? The bot still says `Contributor License Agreement is not agreed yet.`

dhduvall:

aramase (approved):
Thank you for the PR @dhduvall!

Issue #658, opened 2022-12-07
Make the `m.reader.Get` branch return `admission.Allowed` when the SA is not found


aramase:
Should we also make the `m.reader.Get` branch below return `admission.Allowed` when the SA is not found?  Or maybe just keep going with some "default" idea of what the SA state is?

_Originally posted by @enj in https://github.com/Azure/azure-workload-identity/pull/652#discussion_r1040015222_

PR #720, opened 2023-01-25
feat: remove pod/service account labeled check in webhook

aramase:

**Reason for Change**:

With the objectselector set in mwh, only pods that have the label `azure.workload.identity/use: "true"` will be sent to the webhook for mutation. We no longer need the check for pod/service account labeled.

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

fixes https://github.com/Azure/azure-workload-identity/issues/658

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -30,13 +30,6 @@ var (
 	// ProxyImageVersion is the image version of the proxy init and sidecar.
 	// This is injected via LDFLAGS in the Makefile during the build.
 	ProxyImageVersion string
-
-	PodLabelMissingWarning = fmt.Sprintf(`pod missing label '%s: "true"'. This label will be required in a future release for the webhook to mutate pod. Please add the label to the pod.`, UseWorkloadIdentityLabel)
-)
-
-const (
-	// warningAnnotationKey is the annotation key used to store warning messages in audit annotations
-	warningAnnotationKey = "mutation.azure-workload-identity.io/warning"
 )
 
 // +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create,versions=v1,name=mutation.azure-workload-identity.io,sideEffects=None,admissionReviewVersions=v1;v1beta1,matchPolicy=Equivalent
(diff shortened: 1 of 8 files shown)


pkg/webhook/webhook.go
```diff
@@ -89,8 +82,6 @@ func NewPodMutator(client client.Client, reader client.Reader, arcCluster bool,
 
 // PodMutator adds projected service account volume for incoming pods if service account is annotated
 func (m *podMutator) Handle(ctx context.Context, req admission.Request) (response admission.Response) {
-	warnings := []string{}
```
enj, line 92: Can we remove the `syncBuffer` from the e2e test, I think that was forgotten in the last PR.

pkg/webhook/webhook.go
```diff
@@ -125,39 +116,15 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
...
-	}
-	if !podLabeled {
-		warnings = append(warnings, PodLabelMissingWarning)
-	}
-
-	// only log metrics for the pods that are actually mutated
 	defer func() {
 		if m.reporter != nil {
 			m.reporter.ReportRequest(ctx, req.Namespace, time.Since(timeStart))
 		}
```
enj, line 127: Can we move this `defer` up?

pkg/webhook/webhook.go
```diff
@@ -125,39 +116,15 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
 			return admission.Errored(http.StatusBadRequest, err)
 		}
 		// bypass cache and get from the API server as it's not found in cache
-		err = m.reader.Get(ctx, types.NamespacedName{Name: serviceAccountName, Namespace: pod.Namespace}, serviceAccount)
-		if err != nil {
-			logger.Error(err, "failed to get service account")
-			return admission.Errored(http.StatusBadRequest, err)
+		if err = m.reader.Get(ctx, types.NamespacedName{Name: serviceAccountName, Namespace: pod.Namespace}, serviceAccount); err != nil {
+			logger.Error(err, "failed to get service account, using defaults")
```
enj, line 120: Should we remove this logic from the above cache'd get as well?

enj:
> client id env var will be empty

@aramase what exactly does that _mean_ for the app?

pkg/webhook/webhook.go
aramase: In case of cache'd get we only return error if it's not `apierrors.IsNotFound` https://github.com/Azure/azure-workload-identity/blob/0ace9ecb69f10438ad6abe91970a8859d6e2cb60/pkg/webhook/webhook.go#L122-L125

For not found error, we try doing a get from API server and with this change even if that doesn't exist we just proceed without failing.

aramase:
> > client id env var will be empty
> 
> @aramase what exactly does that _mean_ for the app?

If the app relies on `AZURE_CLIENT_ID` env var, then the token request will fail. 

Either the service account is created simultaneously when the pod is created and we don't see it while mutating or the service account doesn't exist in which case pod will fail with service account not found?

pkg/webhook/webhook.go
enj: Right I was more saying "maybe just ignore all errors here too just like in the live look up case"

aramase: that sounds good!

aramase: updated to ignore errors and also added text about `AZURE_CLIENT_ID` being empty in the error log.

aramase: removed these changes as discussed!

enj (approved):

PR #749, opened 2023-02-09
Updated docs so that azure.workload.identity/proxy-sidecar-port is de…

ilya-scale:
…fault false

**Reason for Change**:

It seems there is a bug in the documentation regarding sidecar container. The tests seem to expect false when the value is not present, which also makes sense since not everyone will need  it.

no

no

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [x ] no

**Notes for Reviewers**:


aramase (commented):
Thank you for the PR @ilya-scale! The webhook checks if the annotation `azure.workload.identity/inject-proxy-sidecar` exists in the pod and the value used (true or false) should have no impact on the behavior. If you want to inject the proxy side car, you can add the annotation, if not required, then you should remove it from the annotations field.

ilya-scale:
It was a bit confusing though that it says that the default is "true". I at least read it as: by default the sidecar is injected which does not seem to be the case. Maybe there is some better way of communicating this.

aramase:
> It was a bit confusing though that it says that the default is "true". I at least read it as: by default the sidecar is injected which does not seem to be the case. Maybe there is some better way of communicating this.

Agreed! Thanks for opening the PR. I'm going to close this and open a follow-up to add a separate column to document the default behavior.

PR #782, opened 2023-03-07
fix: remove unnecessary rbac permissions for mwh

aramase:
**Reason for Change**:

- Remove `CREATE`, `PATCH`, `UPDATE` for service accounts
- Remove `CREATE`, `PATCH`  and `DELETE` for mutating webhook configuration

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -11,22 +11,16 @@ rules:
   resources:
   - serviceaccounts
   verbs:
-  - create
   - get
   - list
-  - patch
-  - update
   - watch
 - apiGroups:
   - admissionregistration.k8s.io
   resources:
   - mutatingwebhookconfigurations
   verbs:
-  - create
-  - delete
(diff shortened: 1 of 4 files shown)


config/rbac/role.yaml
```diff
@@ -11,19 +11,14 @@ rules:
...
 - apiGroups:
   - admissionregistration.k8s.io
   resources:
   - mutatingwebhookconfigurations
   verbs:
-  - create
-  - delete
   - get
   - list
   - patch
```
enj, line 24: Do we need patch and update?

aramase: Just `UPDATE` is required. I've updated the PR!

sozercan (approved):

PR #794, opened 2023-03-15
feat: set `reinvocationPolicy: IfNeeded` for webhook

### LINKED ISSUE #787: Admission webhook to use `reinvocationPolicy: IfNeeded` because of unknown admissions webhook ordering
I'm one of the maintainers of Dapr (https://dapr.io) which is available as a managed extension for AKS and for ACA.

Dapr has a mutating admission webhook which modifies the container spec of a deployment object to add the Dapr sidecar as an additional container.

The problem with mutating admission webhooks is that the ordering is not guaranteed. It is critical that the environment variables injected by the Workload Identity mutating admissions webhook also get injected into the Dapr sidecar.

Therefore, can you please advice whether you are specifying the "reinvocationPolicy: IfNeeded" option for the Workload Identity webhook? This would retrigger the webhook in the case that the Workload Identity webhook was applied first, then Dapr's sidecar injector webhook got triggered second. What we would want to see in this case is that the Workload Identity Admissions Webhook gets triggered again.

Please see https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy

### LINKED ISSUE #791: Workload Identity does not work with Dapr Sidecar (or any injected container)
**Describe the bug**

The Dapr sidecar is unable to authenticate with Workload Identity. In fact this likely applies to every injected container that is injected via a mutating admissions webhook.

I performed my tests in AKS only - so cannot speak for another environment. The AKS workload identity in my tests always triggered first before the Dapr sidecar container was injected into the pod. But the workload identity webhook doesn't retrigger (see #787 needeed to address this). As such the Dapr sidecar does not receive the four required workload identity environment variables and the token mount.

**Steps To Reproduce**

This currently requires a manual build of Dapr - but any injected sidecar (via a mutating admissions webhook) will have the same problem. Simply running `kubectl describe pod ...` will be sufficient to see that environment variables were not injected into the sidecar.

**Expected behavior**

Workload Identity environment variables and token are injected into the sidecar container.

aramase:
**Reason for Change**:

- Sets `reinvocationPolicy: IfNeeded` for the workload identity mutating webhook.

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

fixes #787
fixes #791 

- [ ] yes
- [ ] no

**Notes for Reviewers**:
- Validated this with the [dapr](https://learn.microsoft.com/en-us/azure/aks/dapr) extension.

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -16,6 +16,7 @@ webhooks:
   failurePolicy: Fail
   matchPolicy: Equivalent
   name: mutation.azure-workload-identity.io
+  reinvocationPolicy: IfNeeded
   rules:
   - apiGroups:
     - ""
--- a/manifest_staging/charts/workload-identity-webhook/templates/azure-wi-webhook-mutating-webhook-configuration-mutatingwebhookconfiguration.yaml
+++ b/manifest_staging/charts/workload-identity-webhook/templates/azure-wi-webhook-mutating-webhook-configuration-mutatingwebhookconfiguration.yaml
@@ -25,6 +25,7 @@ webhooks:
   objectSelector:
     matchLabels:
       azure.workload.identity/use: "true"
+  reinvocationPolicy: IfNeeded
   rules:
   - apiGroups:
     - ""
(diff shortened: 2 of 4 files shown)


sozercan (approved):

berndverst:
Thanks for the quick turnaround!

Issue #824, opened 2023-04-05
azure-wi-webhook-controller-manager brokes aks upgrade because of pdb


mac2000:
Our attempt to upgrade AKS failed after a hour and put whole cluster in failed state 🤷‍♂️ 

```
Failed to save Kubernetes service 'capybara002-aks-eun-dev'.

Error: Drain of aks-infra002-17497433-vmss00000y did not complete:

Too many req pod azure-wi-webhook-controller-manager-749b9d5cc8-2chl9 on node aks-infra002-17497433-vmss00000y:

azure-workload-identity-system/azure-wi-webhook-controller-manager-749b9d5cc8-2chl9 blocked by pdb azure-wi-webhook-controller-manager with unready pods [].

See http://aka.ms/aks/debugdrainfailures
```

Imagine how bad and scarry it was :) thankfully it was not production but development

In AKS upgrade [docs](https://learn.microsoft.com/en-us/azure/aks/upgrade-cluster) there is mention about this:

> Ensure that any PodDisruptionBudgets (PDBs) allow for at least 1 pod replica to be moved at a time otherwise the drain/evict operation will fail. If the drain operation fails, the upgrade operation will fail by design to ensure that the applications are not disrupted. Please correct what caused the operation to stop (incorrect PDBs, lack of quota, and so on) and re-try the operation.

But this budgets are hidden inside chart and not configurable and/or mentioned anywhere in docs (to be fair I did not even know that we have one in our cluster)

As a workaround I have manually replaced `minAvailable` to `maxUnavailable` and with help of az cli rerun node pool upgrade and then whole cluster - it helped and cluster healthy again 

**The problem here:** if we leave this as is - it will mean that we can not upgrade, and even worse, enable auto upgrade because it will be always failing exactly because of this deployment

So I am wondering if there some other workarounds we may use or may be at least make this part of chart configurable

mac2000:
Meanwhile the workaround is

before making an upgrade run

```bash
kubectl -n azure-workload-identity-system patch pdb azure-wi-webhook-controller-manager --type=json -p='[{"op": "remove", "path": "/spec/minAvailable"},{"op": "add", "path": "/spec/maxUnavailable", "value": 1}]'
```

after upgrade

```bash
[...]

Issue #826, opened 2023-04-06
Use `maxUnavailable: 1` instead of `minAvailable` in pod disruption budget


aramase:

enj:
Closing because the current default of 2 replicas works correctly with `minAvailable=1`.

PR #827, opened 2023-04-06
feat: make podDisruptionBudget minAvailable/maxUnavailable configurable

aramase:
- make `minAvailable` and `maxUnavailable` configurable in helm charts
- opened https://github.com/Azure/azure-workload-identity/issues/826 to track setting `maxUnavailable` instead of `minAvailable`. 

fixes https://github.com/Azure/azure-workload-identity/issues/824


enj (approved):
This LGTM but how exactly do we test that it works?

aramase:
> This LGTM but how exactly do we test that it works?

With this PR, we're only changing the default value to be customizable. We can try to add a `drain` test (if _possible_) with https://github.com/Azure/azure-workload-identity/issues/826 to validate this behavior.

PR #828, opened 2023-04-06
chore: update golangci-lint to v1.52.2

aramase:
- update golangci-lint to `v1.52.2`


aramase:
This change was done in https://github.com/Azure/azure-workload-identity/pull/832.

PR #867, opened 2023-04-25
docs: add note about many:1 and 1:many in quick start

aramase:
- Add a note about mapping multiple Kubernetes service account to single identity and multiple identities with single Kubernetes service account in FAQ. 
- Also add a reference note in quick start.


sozercan (approved):
Thanks! LGTM

PR #869, opened 2023-04-25
docs: use mermaid diagram in introduction

aramase:
- Replace the sequence diagram image in introduction with mermaid diagram


sozercan (approved):
LGTM

PR #946, opened 2023-05-09
chore: update distroless-iptables to v0.2.4

aramase:
- update distroless-iptables to v0.2.4

### DIFF
--- a/docker/proxy-init.Dockerfile
+++ b/docker/proxy-init.Dockerfile
@@ -1,4 +1,4 @@
-ARG BASEIMAGE=registry.k8s.io/build-image/distroless-iptables:v0.2.3
+ARG BASEIMAGE=registry.k8s.io/build-image/distroless-iptables:v0.2.4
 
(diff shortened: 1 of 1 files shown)


docker/proxy-init.Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG BASEIMAGE=registry.k8s.io/build-image/distroless-iptables:v0.2.3
+ARG BASEIMAGE=registry.k8s.io/build-image/distroless-iptables:v0.2.4
```
sozercan, line 1: wish we could automate this. should we just remove the arg so dependabot can bump, wdyt?

aramase: sgtm! pushed https://github.com/Azure/azure-workload-identity/pull/946/commits/729b15877ec5a4591fe02d46206fb893d54ad15d to remove the `ARG`.

sozercan (approved):

PR #1032, opened 2023-07-19
ci: add dependabot group for gomod (part 1)

aramase:
Adding groups for k8s and azure-sdk-for-go deps, so we have a single PR for each (xref: https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/)


sozercan (approved):
Love this! LGTM

PR #1065, opened 2023-08-14
feat: add nightly copa scan and patch workflow

ashnamehrotra:
**Reason for Change**:

Adds a patch workflow to run nightly using the [Copa Github Action](https://github.com/project-copacetic/copa-action) to patch ghcr.io/azure/azure-workload-identity/proxy-init:latest if there are vulnerabilities.

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:


.pipelines/templates/patch-images.yaml
```diff
@@ -0,0 +1,61 @@
+pr: none
```
sozercan, line 1: @ashnamehrotra this is an ADO pipeline, GHA won't work here since copa is not an ADO task

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,58 @@
...
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Push patched image
+        run: |
+          docker tag ghcr.io/azure/azure-workload-identity/proxy-init:patched ${{ matrix.images }}
+          docker push ${{ matrix.images }}
+  create-updated-manifest:
```
sozercan, line 52: do we need `needs: patch` here?

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,58 @@
...
+jobs:    
+  patch: 
+      runs-on: ubuntu-latest
+      strategy:
+        fail-fast: false
+        matrix:
+          images: ['ghcr.io/azure/azure-workload-identity/proxy-init:latest-linux-arm64', 'ghcr.io/azure/azure-workload-identity/proxy-init:latest-linux-amd64']
+      steps:
+      - name: Generate Trivy Report
+        uses: aquasecurity/trivy-action@41f05d9ecffa2ed3f1580af306000f734b733e54
```
sozercan, line 17: nit: add a comment with the version

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Check Vuln Count
+        id: vuln_cout
+        run: |
+          report_file="report.json"
+          vuln_count=$(jq '.Results | length' "$report_file")
+          echo "vuln_count=$vuln_count" >> $GITHUB_OUTPUT
+      - name: Copa Action
+        if: steps.vuln_cout.outputs.vuln_count != '0'
+        id: copa
+        uses: project-copacetic/copa-action@v1.0.0
```
sozercan, line 34: copacetic-action now i think?

please add hash and version too

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,58 @@
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  pull_request:
+    branches:
+      - main
+
```
sozercan, line 7: add permission, i think this'll need read for content and write for packages

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,58 @@
...
+        if: steps.vuln_cout.outputs.vuln_count != '0'
+        id: copa
+        uses: project-copacetic/copa-action@v1.0.0
+        with:
+          image:  ${{ matrix.images }}
+          image-report: 'report.json'
+          patched-tag: 'patched'
+          buildkit-version: 'v0.12.1'
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@ecf95283f03858871ff00b787d79c419715afc34
```
sozercan, line 41: do we need the buildx action

sozercan:
@ashnamehrotra can you make sure to test this in your fork?

.github/workflows/patch-images.yaml
ashnamehrotra: I think it needs to be referenced by the repo name instead of action name

.github/workflows/patch-images.yaml
ashnamehrotra: removed!

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,58 @@
...
+          vuln-type: 'os'
+          image-ref:  ${{ matrix.images }}
+      - name: Check Vuln Count
+        id: vuln_cout
+        run: |
+          report_file="report.json"
+          vuln_count=$(jq '.Results | length' "$report_file")
+          echo "vuln_count=$vuln_count" >> $GITHUB_OUTPUT
+      - name: Copa Action
+        if: steps.vuln_cout.outputs.vuln_count != '0'
```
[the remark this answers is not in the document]
ashnamehrotra: we can but I think then it would need to exit with an error?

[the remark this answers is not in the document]
sozercan: we'll need to exit with 0 exit code if there are no vulns. see here as an example: https://github.com/orgs/community/discussions/46992

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - cron: "0 0 * * *"
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
```
sozercan, line 12:
```suggestion
jobs:
```

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,63 @@
...
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:    
+  patch:
```
sozercan, line 13:
```suggestion
  patch:
```

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,63 @@
+on:
+  schedule:
+    - cron: "0 0 * * *"
```
sozercan, line 3: can we add a comment with the schedule so it's more readable. is this daily?

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,63 @@
...
+        if: steps.copa.conclusion == 'success'
+        run: |
+          docker tag ghcr.io/azure/azure-workload-identity/proxy-init:patched ${{ matrix.images }}
+          docker push ${{ matrix.images }}
+  create-updated-manifest:
+      needs: patch
+      runs-on: ubuntu-latest
+      steps:
+      - name: Login to ghcr.io
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
```
sozercan, line 54:
```suggestion
        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2.2.0
```

.github/workflows/patch-images.yaml
ashnamehrotra: yes this is daily at midnight

.github/workflows/patch-images.yaml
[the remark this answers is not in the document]
ashnamehrotra: tested with exit code of 0, but it looks like it would still continue with the rest of the steps since it would be seen as successful

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,69 @@
+on:
+  schedule:
+    - cron: "0 0 * * *" # nightly
+  pull_request:
+    branches:
+      - main
```
aramase, line 6: why do we need this to be run as part of every PR?

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,69 @@
...
+  packages: write
+
+jobs:
+  patch:
+      runs-on: ubuntu-latest
+      strategy:
+        fail-fast: false
+        matrix:
+          images: ['ghcr.io/azure/azure-workload-identity/proxy-init:latest-linux-arm64', 'ghcr.io/azure/azure-workload-identity/proxy-init:latest-linux-amd64']
+      steps:
```
aramase, line 19: Could you add the `Harden runner` step

https://github.com/Azure/azure-workload-identity/blob/eb7cef39ec3bae6a49f02c90a3fbb257b156f7c6/.github/workflows/publish-images.yaml#L38-L41

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,69 @@
+on:
```
aramase, line 1: add a name to the workflow

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,69 @@
...
+          image:  ${{ matrix.images }}
+          image-report: 'report.json'
+          patched-tag: 'patched'
+          buildkit-version: 'v0.12.1'
+      - name: Push patched image
+        if: steps.copa.conclusion == 'success'
+        run: |
+          docker tag ghcr.io/azure/azure-workload-identity/proxy-init:patched ${{ matrix.images }}
+          docker push ${{ matrix.images }}
+  create-updated-manifest:
```
aramase, line 55:
```suggestion

  create-updated-manifest:
```

aramase (changes requested):

.github/workflows/patch-images.yaml
ashnamehrotra: we don't, removed!

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,73 @@
...
+        if: steps.copa.conclusion == 'success'
+        run: |
+          docker tag ghcr.io/azure/azure-workload-identity/proxy-init:patched ${{ matrix.images }}
+          docker push ${{ matrix.images }}
+
+  create-updated-manifest:
+      needs: patch
+      runs-on: ubuntu-latest
+      steps:
+      - name: Login to ghcr.io
```
aramase, line 63: Could you add the `Harden runner` step here too?

https://github.com/Azure/azure-workload-identity/blob/eb7cef39ec3bae6a49f02c90a3fbb257b156f7c6/.github/workflows/publish-images.yaml#L38-L41

aramase (commented):
Just 1 minor comment, otherwise lgtm!

.github/workflows/patch-images.yaml
```diff
@@ -0,0 +1,77 @@
...
+        uses: project-copacetic/copa-action@1eb86b0907bce48225b66dc9488c7d329c2d48a0 # v1.0.0
+        with:
+          image:  ${{ matrix.images }}
+          image-report: 'report.json'
+          patched-tag: 'patched'
+          buildkit-version: 'v0.12.1'
+      - name: Push patched image
+        if: steps.copa.conclusion == 'success'
+        run: |
+          docker tag ghcr.io/azure/azure-workload-identity/proxy-init:patched ${{ matrix.images }}
```
aramase, line 56: Why are we tagging the images with `:patched` tag here if it's never pushed?

ashnamehrotra: we are tagging the resulting ':patched' image with the actual name of the image so we can push the newer version, since we would want to make sure the new latest tag is the one available in ghcr

aramase: that makes sense! It's been a while since I directly interacted with docker and just confirmed the first arg is `SOURCE_IMG` and the 2nd arg is `TARGET_IMG`. 

Thanks for clarifying!

PR #1066, opened 2023-08-15
ci: add dependabot group for `examples/*` dir (part 2)

aramase:
Adding groups for all deps for a package-ecosystem in `examples/*`, so we have a single PR for each


sozercan (approved):
thank you! lgtm

PR #1090, opened 2023-08-23
fix: disallow injecting proxy sidecar in pods with `hostNetwork: true`

aramase:
- disallow injecting proxy sidecar in pods with `hostNetwork: true`
- add a note about it in known issues


sozercan (approved):
LGTM

PR #1102, opened 2023-08-31
fix: Prepend proxy sidecar container to prevent application container restarts

### LINKED ISSUE #1101: Prepend proxy sidecar to the container list to prevent application restarts
[Currently](https://github.com/Azure/azure-workload-identity/blob/37dc12fdf67ec0bd54d3e3f5c24273a0207707e8/pkg/webhook/webhook.go#L236C7-L236C7), proxy sidecar gets added to the end of the containers list. In this case, the application container may crash a couple of times before the sidecar becomes available.

I suggest prepending the sidecar to the containers list (by swapping containers and corev1.Container{...} in append call), which will block all other containers due to the post-start hook in the sidecar.

Sheludchenko:
**Reason for Change**:
To fix application container startup errors due to incorrect container ordering (proxy sidecar will be triggered last instead of first).

**Issue Fixed**:
Fixes #1101

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -233,7 +233,7 @@ func (m *podMutator) injectProxySidecarContainer(containers []corev1.Container, 
 	}
 
 	logLevel := currentLogLevel() // run the proxy at the same log level as the webhook
-	containers = append(containers, corev1.Container{
+	containers = append([]corev1.Container{{
 		Name:            ProxySidecarContainerName,
 		Image:           strings.Join([]string{imageRepository, ProxyImageVersion}, ":"),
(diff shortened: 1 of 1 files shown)


aramase (changes requested):
Thanks for the PR!

https://github.com/Azure/azure-workload-identity/blob/37dc12fdf67ec0bd54d3e3f5c24273a0207707e8/pkg/webhook/webhook_test.go#L1048 needs to be modified to check the ordering and tests to pass.

aramase:
Thank you for the PR! This comment has been included in https://github.com/Azure/azure-workload-identity/pull/1108.

PR #1146, opened 2023-10-20
chore: Update .NET sample to use MSAL best practices

bgavrilMS:
**Reason for Change**:

The MSAL samples do not showcase MSAL best practices such as token caching, authority configuration and assertion refresh.

Source: I am the engineering manager for MSAL.NET

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:


examples/msal-net/akvdotnet/TokenCredential.cs
```diff
@@ -22,9 +22,12 @@ public MyClientAssertionCredential()
         var tokenPath = Environment.GetEnvironmentVariable("AZURE_FEDERATED_TOKEN_FILE");
         var tenantID = Environment.GetEnvironmentVariable("AZURE_TENANT_ID");
 
-        _confidentialClientApp = ConfidentialClientApplicationBuilder.Create(clientID)
-                .WithClientAssertion(ReadJWTFromFS(tokenPath))
-                .WithTenantId(tenantID).Build();
+        _confidentialClientApp = ConfidentialClientApplicationBuilder
+                .Create(clientID)
+                .WithAuthority("https://login.microsoftonline.com/", tenantID) // for simplicity, this samples assume the app is in the public cloud
```
aramase, line 27: The value for authority is available in` AZURE_AUTHORITY_HOST` environment variable.

examples/msal-net/akvdotnet/TokenCredential.cs
```diff
@@ -22,9 +22,12 @@ public MyClientAssertionCredential()
...
+                .WithClientAssertion(() => ReadJWTFromFS(tokenPath))          // make sure that ReadJWTFromFS always returns a fresh JWT
```
aramase, line 28: Can we do something similar to https://github.com/Azure/azure-workload-identity/blob/ffb15ec83c68d13b5343871f8c404012a45eadc7/examples/msal-go/token_credential.go#L66

Read from file system every 5 minutes instead of every time?

examples/msal-net/akvdotnet/TokenCredential.cs
```diff
@@ -55,6 +58,7 @@ public override async ValueTask<AccessToken> GetTokenAsync(TokenRequestContext r
         return new AccessToken(result.AccessToken, result.ExpiresOn);
     }
 
+    // This assumes the file is always updated with a fresh JWT
```
aramase, line 61: The token lifetime is between 1-24h, so it's not always updated.

aramase (changes requested):
@bgavrilMS Thank you for the PR! Added few comments.

examples/msal-net/akvdotnet/TokenCredential.cs
pmaytak, line 61:
```suggestion
    // This assumes the file always contains an unexpired JWT 
```

examples/msal-net/akvdotnet/TokenCredential.cs
pmaytak, line 28:
```suggestion
                .WithClientAssertion(() => ReadJWTFromFS(tokenPath))          // make sure that ReadJWTFromFS always returns an unexpired JWT 
```

bgavrilMS:
Thanks for the ping @aramase, taking a look.

examples/msal-net/akvdotnet/TokenCredential.cs
bgavrilMS: Yes, done.

bgavrilMS: Yes, done

examples/msal-net/akvdotnet/TokenCredential.cs
bgavrilMS: Ok, as long as there is some buffer time it should be fine. In MSAL we take 5 min buffer.

aramase (approved):
@bgavrilMS Thank you for the PR! 🎉 

LGTM

PR #1186, opened 2023-11-27
Updated the quick start guide to include details on --enable-workload…

akhiljimmy7:
…-identity flag

https://learn.microsoft.com/en-us/azure/aks/workload-identity-deploy-cluster#update-an-existing-aks-cluster

**Reason for Change**:

If an user enabled the workload identity on the AKS Cluster, there is no need of manual installation of mutating admission webhook, By enabling workload identity on the AKS Cluster it installs the mutating admission webhook components in the kube-system namespace by default.

 No

 No

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/docs/book/src/quick-start.md
+++ b/docs/book/src/quick-start.md
@@ -18,7 +18,8 @@ Before we get started, ensure the following:
 ## 1. Complete the installation guide
 
 [Installation guide][13]. At this point, you should have already:
-- installed the mutating admission webhook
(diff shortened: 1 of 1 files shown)


superherointj:
@akhiljimmy7 I think, you should rebase and squash commits.

PR #1193, opened 2023-12-04
chore: Update akvdotnet.csproj to use recommended .NET

bgavrilMS:
**Reason for Change**:

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/examples/msal-net/akvdotnet/akvdotnet.csproj
+++ b/examples/msal-net/akvdotnet/akvdotnet.csproj
@@ -2,11 +2,11 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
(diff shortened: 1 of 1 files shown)


aramase (changes requested):
We need to update these as well?

https://github.com/Azure/azure-workload-identity/blob/0fb935f481b779a0c89724a9f4fc4fb3f221bcf0/examples/msal-net/akvdotnet/Dockerfile#L7-L9

bgavrilMS:
I don't have the capacity to run your builds etc. But be aware that net5 is deprecated.

PR #1208, opened 2023-12-12
chore: use base images from mcr

aramase:
- use base images from mcr
- bump distroless-iptables to `v0.4.3`

### DIFF
--- a/docker/proxy.Dockerfile
+++ b/docker/proxy.Dockerfile
@@ -1,7 +1,4 @@
-ARG BUILDER=mcr.microsoft.com/oss/go/microsoft/golang:1.21-bullseye
-ARG BASEIMAGE=gcr.io/distroless/static:nonroot
-
-FROM ${BUILDER} as builder
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.21-bookworm as builder
 
 ARG LDFLAGS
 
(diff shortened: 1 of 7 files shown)


docker/webhook.Dockerfile
```diff
@@ -1,5 +1,5 @@
-ARG BUILDER=mcr.microsoft.com/oss/go/microsoft/golang:1.21-bullseye
```
sozercan, line 1: are we using these as args anywhere? if not, let's add them directly and have dependabot manage these (applies to all)

aramase: done!

docker/proxy.Dockerfile
```diff
@@ -1,7 +1,6 @@
-ARG BUILDER=mcr.microsoft.com/oss/go/microsoft/golang:1.21-bullseye
-ARG BASEIMAGE=gcr.io/distroless/static:nonroot
+ARG BASEIMAGE=mcr.microsoft.com/cbl-mariner/distroless/minimal:2.0-nonroot
```
sozercan, line 1: let's add it for this too so dependabot can pin to digest

examples/msal-python/Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG BUILDER=debian:11-slim
-ARG BASEIMAGE=gcr.io/distroless/python3-debian11
+ARG BASEIMAGE=mcr.microsoft.com/cbl-mariner/distroless/python:3.9
```
sozercan, line 2: same here

examples/msal-python/Dockerfile
```diff
@@ -1,8 +1,4 @@
-ARG BUILDER=debian:11-slim
-ARG BASEIMAGE=gcr.io/distroless/python3-debian11
-
-# ref: https://github.com/GoogleContainerTools/distroless/blob/main/examples/python3-requirements/Dockerfile
-FROM ${BUILDER}  AS build
+FROM debian:11-slim  AS build
```
sozercan, line 1:
```suggestion
FROM mcr.microsoft.com/mirror/docker/library/debian:bookworm-slim AS build
```

sozercan (approved):

Issue #1210, opened 2023-12-18
azure.workload.identity/inject-proxy-sidecar set to false is not respected


dnitsch:
**Describe the bug**
When setting the annotation of `azure.workload.identity/inject-proxy-sidecar`  to false this is not respected as the `shouldInjectProxySidecar` helper method only checks the existance of the key in the map and not the value. 

The docs do say somewhat misleadingly to set this value to `true` or `false` if you want the migration sidecar container 

https://learn.microsoft.com/en-us/azure/aks/workload-identity-migrate-from-pod-identity#deploy-the-workload-with-migration-sidecar

However, it would be a much nicer user experience to have a flag respected.

**Steps To Reproduce**

create a deployment and set the `azure.workload.identity/inject-proxy-sidecar` to false and then query the pods for that deployment and you will see a `azwi-proxy` and `awei-proxy-init` containers added as well as all the required volumes and env variables.

**Expected behavior**

setting the `azure.workload.identity/inject-proxy-sidecar` value to false is respected and sidecar creation is skipped.

**Logs**

**Environment**

- Kubernetes version (use `kubectl version`):
v1.26.6
GoVersion:"go1.19.10", Compiler:"gc", Platform:"linux/amd64"}
- Cloud provider or hardware configuration:
- Azure (AKS)
- OS (e.g: `cat /etc/os-release`):
- 
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

**Additional context**

PR #1211, opened 2023-12-18
fix: check the value of the annotation to actually create the sidecar

dnitsch:
**Reason for Change**:

#1210 

Ensures annotation values are respected

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


enj:
This is a breaking change since the existing behavior only requires that the annotation be set at all, not the specific value it has.  At most, we can update the documentation to describe the behavior.

dnitsch:
hi @enj, thanks for coming back on this. 

Sure that makes sense and a doc update would be good (happy to PR that, if you could send me the link to the source repo)

Though, the object validation would fail if there is no (string) value for a given key on apply, right?

Possibly, adding an empty `""` to the `positiveAnnotationVal` slice to cover this scenario. 

Again I totally see where you are coming from about the breaking change - it's just that I can't see many people setting that value to `false` and actually expect a sidecar to be injected 😄.

PR #1292, opened 2024-03-18
chore: bump k8s deps to v1.29.4 and controller-runtime to v0.17.3

aramase:
- bump k8s deps to v1.29.4 and controller-runtime to v0.17.3

fixes https://github.com/Azure/azure-workload-identity/issues/1312


aramase:
I've added a unit test in https://github.com/Azure/azure-workload-identity/pull/1292/commits/9640c5881222828a1d9decf0119bcad0222a5c7b to assert the existing failure and the bump of deps fix it.

```bash
=== RUN   TestHandle/pod_has_the_required_label,_restart_policy_in_init_container
    webhook_test.go:806: expected no remove patches, got: {remove /spec/initContainers/0/restartPolicy <nil>}
```

cmd/webhook/main.go
```diff
@@ -190,3 +198,18 @@ func setupProbeEndpoints(mgr ctrl.Manager, setupFinished chan struct{}) {
 	}
 	entryLog.Info("added healthz and readyz check")
 }
+
+func parseTLSVersion(tlsVersion string) (uint16, error) {
+	switch tlsVersion {
+	case "", "1.0":
```
enj, line 204: Empty string should error.

enj (changes requested):

cmd/webhook/main.go
aramase: done!

enj (approved):

Issue #1312, opened 2024-04-03
Kubernetes native sidecar not supported


YvesZelros:
**Describe the bug**

AWI remove the field **_restartPolicy_** on init containers that break the support of [Kubernetes native sidecar](https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/)

**Steps To Reproduce**

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: simple-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: simple-deployment
  template:
    metadata:
      labels:
        app: simple-deployment
        azure.workload.identity/use: "true"
    spec:
      initContainers:
        - command:
            - "/bin/sh"
            - "-c"
            - "while true; do echo 'hello from init'; sleep 1; done"
          image: busybox
          name: init
          restartPolicy: Always
      containers:
        - command:
            - "/bin/sh"
            - "-c"
            - "while true; do echo 'hello from main'; sleep 1; done"
          image: busybox
          name: main
```
**Expected behavior**

The main container must be running and not waiting the end of the init container as the init container is a native sidecar (restartPolicy = Always).

**Logs**
NAME                                 READY   STATUS     RESTARTS   AGE
simple-deployment-7854b77d4b-hbs7r   0/1     Init:0/1   0          3m35s

**Environment**
- Kubernetes version (use `kubectl version`): Server Version: v1.29.2
- Cloud provider or hardware configuration: AKS

**Additional context**
https://github.com/istio/istio/issues/49221

aramase:
Thanks for reporting the issue. I have a PR open to update the deps: https://github.com/Azure/azure-workload-identity/pull/1292.

eplightning:
Any chance of getting a new release soon and deploying that on AKS? This is blocking us from using new sidecar functionality.

mikec-greene:
Is it possible to get an ETA on when we might see this appearing in AKS? We're also blocked from using native sidecars in Istio, many thanks.

rene-bos:
The PR which fixes this issue (#1292) has already been released in [v1.3.0](https://github.com/Azure/azure-workload-identity/releases/tag/v1.3.0) on the 6th of June. So you can already install it in your AKS cluster 😄.

mikec-greene:
[...]

PR #1333, opened 2024-04-30
fix: Filter patches to drop `"remove"` operation to prevent dropping unknown fields

aramase:
**Reason for Change**:

Filter patches to only allow "add" operation to prevent dropping unknown fields

**Issue Fixed**:

fixes https://github.com/Azure/azure-workload-identity/issues/1312

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/pkg/errors"
+	"gomodules.xyz/jsonpatch/v2"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
@@ -166,7 +167,21 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
 		logger.Error("failed to marshal pod object", err)
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+
+	patchResponse := admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+	patches := []jsonpatch.JsonPatchOperation{}
+	for _, patch := range patchResponse.Patches {
+		// drop patches that are removing fields
+		// this prevents patches that remove fields that the webhook doesn't know about
+		// ref: https://github.com/Azure/azure-workload-identity/issues/1312
+		if patch.Operation == "remove" {
+			continue
+		}
+		patches = append(patches, patch)
+	}
+
+	patchResponse.Patches = patches
+	return patchResponse
 }
 
 // PodMutator implements admission.DecoderInjector
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -12,10 +12,10 @@ import (
 
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/pkg/errors"
-	"gomodules.xyz/jsonpatch/v2"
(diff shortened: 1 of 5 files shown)


pkg/webhook/webhook.go
```diff
@@ -166,7 +167,30 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
...
 	}
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+
+	// patchResponse := admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+	// patches := []jsonpatch.JsonPatchOperation{}
+	// for _, patch := range patchResponse.Patches {
+	// 	// drop patches that are removing fields
+	// 	// this prevents patches that remove fields that the webhook doesn't know about
+	// 	// ref: https://github.com/Azure/azure-workload-identity/issues/1312
+	// 	if patch.Operation == "remove" {
```
enj, line 177: This commented out approach seems safer?  And maybe invert the check so we only keep the type of operations we understand?

pkg/webhook/webhook.go
```diff
@@ -166,7 +167,30 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
...
+	// 		continue
+	// 	}
+	// 	patches = append(patches, patch)
+	// }
+	//
+	// patchResponse.Patches = patches
+	// return patchResponse
+
+	// Merge the original raw object with the modified marshaled pod
+	patchedObject, err := strategicpatch.StrategicMergePatch(req.Object.Raw, marshaledPod, corev1.Pod{})
```
enj, line 187: I haven't seen any other webhook use this approach, so it makes me nervous that it won't work correctly in all cases.

pkg/webhook/webhook_test.go
```diff
@@ -728,39 +728,44 @@ func TestHandle(t *testing.T) {
...
-			podLabels: map[string]string{
-				UseWorkloadIdentityLabel: "true",
-			},
+			name:          "pod has the required label, no warnings",
+			rawPod:        newPodRaw("pod", "ns1", "sa", map[string]string{UseWorkloadIdentityLabel: "true"}, nil, false),
+			clientObjects: serviceAccounts,
+			readerObjects: nil,
+		},
+		{
+			name:          "pod has the required label, restart policy in init container",
```
enj, line 767: Also add tests where we fabricate fields that will never exist at different nesting levels that don't exist anywhere in `corev1.Pod`.  Like `pod.Panda` or `pod.Spec.Panda` or `pod.Spec.Jungle.Panda`.

aramase:
Closing this PR after further testing.

In e2e tests we create a pod with init container and test injecting sidecar proxy. The sidecar is added as the first container in the list so it's started first and the way patches get generated instead of doing a complete replace of
`/spec/containers/0`, it does replace and remove on individual fields (which is expected)
 
```yaml
{replace /spec/containers/0/name azwi-proxy}
{replace /spec/containers/0/image aramase/proxy:add_op_00}
{add /spec/containers/0/ports [map[containerPort:8080]]}
{add /spec/containers/0/securityContext/privileged false}
{replace /spec/containers/0/securityContext/runAsNonRoot true}
{add /spec/containers/0/securityContext/readOnlyRootFilesystem true}
{remove /spec/containers/0/securityContext/seccompProfile <nil>}
{remove /spec/containers/0/securityContext/runAsUser <nil>}
{remove /spec/containers/0/args/2 <nil>}
{replace /spec/containers/0/args/1 --log-level=info}
{replace /spec/containers/0/args/0 --proxy-port=8080}
```

if we skip "remove" operation, it breaks because the args and other fields incorrectly remain.

1. The k8s version bump will resolve the current issue with Kubernetes native sidecar fields.
2. I've opened  https://github.com/Azure/azure-workload-identity/issues/1337 for us to investigate further.

PR #1419, opened 2024-07-30
Extend helm chart to create secret token and include automount toggle…

vakaobr:
… for service account

**Reason for Change**:

- Depending of the security controls in place on the cluster, service accounts may have the `automountServiceAccountToken` property set to `false`
- Starting on Kubernetes 1.24, the automatic creation of Secrets of type kubernetes.io/service-account-token was disabled by default. This change introduces this secret to be deployed with the chart in order to it work

**Requirements**

- [X] squashed commits
- [X] included documentation

**Issue Fixed**:

- [ ] yes
- [X] no

**Notes for Reviewers**:

- In current state, when deploying the current chart to a cluster with policies to prevent service account automount, it fails.

### DIFF
--- a/manifest_staging/charts/workload-identity-webhook/templates/azure-wi-webhook-admin-serviceaccount copy.yaml
+++ b/manifest_staging/charts/workload-identity-webhook/templates/azure-wi-webhook-admin-serviceaccount copy.yaml
@@ -1,0 +1,15 @@
+{{- if .Values.imagePullSecrets }}
+imagePullSecrets:
+{{- toYaml .Values.imagePullSecrets | nindent 2 }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


vakaobr:

vakaobr:
Hey @aramase @enj maybe you folks want to have a look on this one? Thanks :)

aramase (changes requested):
Thank you for the PR!

1. You need to update the helm chart generator [here](https://github.com/Azure/azure-workload-identity/tree/main/third_party/open-policy-agent/gatekeeper/helmify) with the changes and then run `make generate`.
2. Also need to add the new variable in the helm chart readme with a description on what it means.

Issue #773, opened 2023-03-01
azure.workload.identity/inject-proxy-sidecar blocks jobs in Kubernetes


dozer75:
**Describe the bug**

We have some jobs and cronjobs that is running in AKS that connects to an Azure SQL database using ODBC. We are planning to use Managed Identity and workload identity to do the authentication in the ODBC driver, and for this we need to use the injection of the proxy sidecar (for some reason). 

But, by doing this, the job won't end after the job container has successfully completed since the sidecar proxy is still alive after our container is done.

The job pod is in state `NotReady` as the proxy container is still running.
```
NAME                                READY   STATUS     RESTARTS   AGE
pod/job-onetime-42mx8   1/2     NotReady   0          26m
```

Here is the dump of the pod:
```
Name:             job-onetime-42mx8
Namespace:        job-jobs
Priority:         0
Service Account:  job-serviceaccount
Node:             aks-systempool-85002938-vmss000002/10.1.0.4
Start Time:       Wed, 01 Mar 2023 16:30:24 +0100
Labels:           azure.workload.identity/use=true
                  controller-uid=9f307dd8-31d6-4c03-b482-563d7fece75e
                  job-name=job-onetime
Annotations:      azure.workload.identity/inject-proxy-sidecar: true
Status:           Running
IP:               10.2.0.19
IPs:
  IP:           10.2.0.19
Controlled By:  Job/job-onetime
Init Containers:
  azwi-proxy-init:
    Container ID:   containerd://148eba569dabca978fb85499303ec4b8de859b6300957962ad5ba9fbf2773008
    Image:          mcr.microsoft.com/oss/azure/workload-identity/proxy-init:v0.15.0
    Image ID:       mcr.microsoft.com/oss/azure/workload-identity/proxy-init@sha256:e8064cf26147bb98efe33c5bc823eb3b32c6b0cbf93619fa6b5d72f4f7a7c068
    Port:           <none>
    Host Port:      <none>
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Wed, 01 Mar 2023 16:30:25 +0100
      Finished:     Wed, 01 Mar 2023 16:30:25 +0100
    Ready:          True
    Restart Count:  0
    Environment:
      PROXY_PORT:                  8000
      AZURE_CLIENT_ID:             
      AZURE_TENANT_ID:             
      AZURE_FEDERATED_TOKEN_FILE:  /var/run/secrets/azure/tokens/azure-identity-token
[...]

PR #1442, opened 2024-08-21
feat: allow proxies to be injected using a native sidecar

devjoes:
**Reason for Change**:

Fixes [Issue 773](https://github.com/Azure/azure-workload-identity/issues/773) by using [native sidecars](https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/) so that pods exit cleanly when running as cronjobs, currently azwi sidecar prevents the pod from terminating. 

It also addresses an annoyance where the first container in the pod is the proxy, so any `kubectl` commands target that container, previously I was working around this by adding the [default-container](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubectl-kubernetes-io-default-container) annotation.

When running on kubernetes >=1.28 it will use native sidecars, otherwise it will use the original approach.

No update to deployment.yaml

No change to Helm chart

**Requirements**

- [X] squashed commits
- [X] included documentation

**Issue Fixed**:

Fixes #733

- [ ] yes
- [X] no

**Notes for Reviewers**:
Tested on v1.30.0-eks-fff26e3

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -20,6 +20,9 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	utilversion "k8s.io/apimachinery/pkg/util/version"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
 	"k8s.io/utils/ptr"
 
 	"github.com/Azure/azure-workload-identity/pkg/config"
@@ -53,10 +56,11 @@ type podMutator struct {
 	azureAuthorityHost string
 	proxyImage         string
 	proxyInitImage     string
+	useNativeSidecar   bool
 }
 
 // NewPodMutator returns a pod mutation handler
-func NewPodMutator(client client.Client, reader client.Reader, audience string, scheme *runtime.Scheme) (admission.Handler, error) {
+func NewPodMutator(client client.Client, reader client.Reader, audience string, scheme *runtime.Scheme, restConfig *rest.Config) (admission.Handler, error) {
 	c, err := config.ParseConfig()
 	if err != nil {
 		return nil, err
@@ -64,6 +68,19 @@ func NewPodMutator(client client.Client, reader client.Reader, audience string, 
 	if audience == "" {
 		audience = DefaultAudience
 	}
+
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(restConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create discovery client")
+	}
+	// "SidecarContainers" went beta in 1.29. With the 3 version skew policy,
+	// between API server and kubelet, 1.32 is the earliest version this can be
+	// safely used.
+	useNativeSidecar, err := serverVersionGTE(discoveryClient, utilversion.MajorMinor(1, 32))
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to check kubernetes version")
+	}
+
 	// this is used to configure the AZURE_AUTHORITY_HOST env var that's
 	// used by the azure sdk
 	azureAuthorityHost, err := getAzureAuthorityHost(c)
@@ -92,6 +109,7 @@ func NewPodMutator(client client.Client, reader client.Reader, audience string, 
 		azureAuthorityHost: azureAuthorityHost,
 		proxyImage:         proxyImage,
 		proxyInitImage:     proxyInitImage,
+		useNativeSidecar:   useNativeSidecar,
 	}, nil
 }
 
@@ -155,7 +173,11 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
 		}
 
 		pod.Spec.InitContainers = m.injectProxyInitContainer(pod.Spec.InitContainers, proxyPort)
-		pod.Spec.Containers = m.injectProxySidecarContainer(pod.Spec.Containers, proxyPort)
+		if m.useNativeSidecar {
+			pod.Spec.InitContainers = m.injectProxySidecarContainer(pod.Spec.InitContainers, proxyPort, ptr.To(corev1.ContainerRestartPolicyAlways))
+		} else {
+			pod.Spec.Containers = m.injectProxySidecarContainer(pod.Spec.Containers, proxyPort, nil)
+		}
 	}
 
 	// get service account token expiration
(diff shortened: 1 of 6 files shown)


devjoes:

devjoes:
@aramase @enj any update on this PR? I've just rebased it

devjoes:
@bench I think that this repo is dead tbh. I'm going to get on to our account managers and see if they can talk to whoever is meant to be in charge of it

docs/book/src/topics/service-account-labels-and-annotations.md
```diff
@@ -21,9 +21,9 @@ All annotations are optional. If the annotation is not specified, the default va
...
 | `azure.workload.identity/inject-proxy-sidecar`             | Injects a proxy init container and proxy sidecar into the pod. The proxy sidecar is used to intercept token requests to IMDS and acquire an AAD token on behalf of the user with federated identity credential.                                                                                                                                                                                                                               | `true`                                    |
+| `azure.workload.identity/use-native-sidecar`               | If injected then the proxy sidecar should be injected as a native sidecar init container instead of a container. Native sidecars are terminated when if pod exits, for instance when running as a cron-job.                                                                                                                                                                                                                                   | `false`                                   |
```
aramase, line 24: Add a note about the minimum required kubernetes version this annotation can be used.

pkg/webhook/webhook.go
```diff
@@ -154,8 +154,14 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
...
 
+		useNativeSidecar := useNativeProxySidecar(pod)
+
 		pod.Spec.InitContainers = m.injectProxyInitContainer(pod.Spec.InitContainers, proxyPort)
-		pod.Spec.Containers = m.injectProxySidecarContainer(pod.Spec.Containers, proxyPort)
+		if useNativeSidecar {
+			pod.Spec.InitContainers = m.injectProxySidecarContainer(pod.Spec.InitContainers, proxyPort, ptr.To(corev1.ContainerRestartPolicyAlways))
+		} else {
+			pod.Spec.Containers = m.injectProxySidecarContainer(pod.Spec.Containers, proxyPort, nil)
+		}
```
aramase, line 164:
```suggestion
                var restartPolicy *corev1.ContainerRestartPolicy
		if useNativeSidecar {
			restartPolicy = ptr.To(corev1.ContainerRestartPolicyAlways)
		}
		pod.Spec.Containers = m.injectProxySidecarContainer(pod.Spec.Containers, proxyPort, restartPolicy)
```

pkg/webhook/webhook.go
```diff
@@ -228,13 +234,18 @@ func (m *podMutator) injectProxyInitContainer(containers []corev1.Container, pro
...
 }
 
-func (m *podMutator) injectProxySidecarContainer(containers []corev1.Container, proxyPort int32) []corev1.Container {
+func (m *podMutator) injectProxySidecarContainer(containers []corev1.Container, proxyPort int32, restartPolicy *corev1.ContainerRestartPolicy) []corev1.Container {
+	imageRepository := strings.Join([]string{ProxyImageRegistry, ProxySidecarImageName}, "/")
 	for _, container := range containers {
 		if container.Name == ProxySidecarContainerName {
-			return containers
+			// proxy-init starts with proxy, so we append ":" to imageRepository when checking
+			if strings.HasPrefix(container.Image, fmt.Sprintf("%s:", imageRepository)) || container.Name == ProxySidecarContainerName {
```
aramase, line 242: this check was removed as part of https://github.com/Azure/azure-workload-identity/pull/1443

pkg/webhook/webhook.go
```diff
@@ -228,13 +234,18 @@ func (m *podMutator) injectProxyInitContainer(containers []corev1.Container, pro
...
+				return containers
+			}
 		}
 	}
 	logLevel := currentLogLevel() // run the proxy at the same log level as the webhook
+
```
aramase, line 248:
```suggestion
```

pkg/webhook/webhook_test.go
```diff
@@ -1151,7 +1151,7 @@ func TestInjectProxySidecarContainer(t *testing.T) {
 	m := &podMutator{proxyImage: imageURL}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			containers := m.injectProxySidecarContainer(test.containers, proxyPort)
+			containers := m.injectProxySidecarContainer(test.containers, proxyPort, nil)
```
aramase, line 1154: Update this table test to include restart policy and ensure the return result contains the desired restart policy

aramase (changes requested):
Please also add a e2e test [here](https://github.com/Azure/azure-workload-identity/blob/main/test/e2e/proxy_test.go) to validate the new annotation.

docs/book/src/topics/service-account-labels-and-annotations.md
enj: This is the wrong approach IMO.  I think we should instead detect at runtime (once at webhook start up time) if the cluster supports native sidecars, and if it does, we should simply always use them for the proxy sidecar.  If for some reason we cannot detect this at runtime, we can make it a CLI flag based config to the webhook (possibly enabled by default since in a few days Kube v1.30 will be the oldest supported upstream version).

devjoes: I've updated the PR to detect the version of k8s used and automatically use a native sidecar if it is supported

pkg/webhook/webhook.go
devjoes: I don't think this will work as native sidecars need to be init containers not containers

pkg/webhook/webhook.go
devjoes: Addressed

pkg/webhook/webhook.go
devjoes: Addressed

pkg/webhook/webhook.go
```diff
@@ -53,17 +55,24 @@ type podMutator struct {
...
+func NewPodMutator(client client.Client, reader client.Reader, audience string, scheme *runtime.Scheme, k8sConfig *rest.Config) (admission.Handler, error) {
 	c, err := config.ParseConfig()
 	if err != nil {
 		return nil, err
 	}
 	if audience == "" {
 		audience = DefaultAudience
 	}
+
+	useNativeSidecar, err := util.AssertK8sMinVersion(k8sConfig, 1, 28)
```
enj, line 71: Looks like `SidecarContainers` went beta in 1.29.  With the 3 version supported skew between API server and kubelet, I think 1.32 is earliest version we can enable this.

```suggestion
	useNativeSidecar, err := util.AssertK8sMinVersion(k8sConfig, 1, 32)
```

devjoes: Good spot, updated

enj (approved):

aramase:
Simulated test failure succesfully: https://github.com/Azure/azure-workload-identity/pull/1588#issuecomment-2787718183

PR #1568, opened 2025-03-11
test: Make a test fail if env var constants change

chlowell:
Downstream consumers such as the Azure SDK depend on these values, so it helps to have a test that would fail if they ever changed.


aramase (approved):
LGTM!

PR #1584, opened 2025-03-31
ci: enable tests with kubernetes v1.32

aramase:
- enable tests with Kubernetes v1.32
- remove 1.28 and 1.29 (EOL)


scripts/create-kind-cluster.sh
```diff
@@ -57,7 +57,7 @@ EOF
 
 create_kind_cluster() {
   # create a kind cluster
-  cat <<EOF | ${KIND} create cluster --name "${KIND_CLUSTER_NAME}" --image "kindest/node:${KIND_NODE_VERSION:-v1.22.4}" --config=-
+  cat <<EOF | ${KIND} create cluster --name "${KIND_CLUSTER_NAME}" --image "mcr.microsoft.com/mirror/kindest/node:${KIND_NODE_VERSION:-v1.32.3}" --config=-
```
aramase, line 60: The CI jobs were getting throttled while pulling images from dockerhub, so switched it to using the mirror in mcr.

enj (approved):

Issue #1555, opened 2025-02-14
Don't add empty AZURE_CLIENT_ID


weisdd:
**Is your feature request related to a problem? Please describe.**

When `AZURE_CLIENT_ID` env is imported into a pod using  `envFrom` `secretRef` method (meaning no annotation is set on a ServiceAccount), the workload identity webhook adds an empty `AZURE_CLIENT_ID` to `env` section of a container spec.
Due to `env` section having higher precedence in Kubernetes, when a container starts it gets an empty `AZURE_CLIENT_ID`.

**Describe the solution you'd like**

Since having empty `AZURE_CLIENT_ID` is of no use, it would be better if webhook stops adding empty `AZURE_CLIENT_ID`.

**Describe alternatives you've considered**

```yaml
env:
  - name: AZURE_CLIENT_ID
    valueFrom:
      secretKeyRef:
        name: postgresql-scripts-database-backups-kv
        key: AZURE_CLIENT_ID
```

The issue with this approach is that not every helm chart offers a way to pass `secretKeyRef`, sometimes only `envFrom` `secretRef` is available.

**Additional context**

In our environment, terraform is used to generate a managed identity and to save a client ID into an Azure Key Vault.

Then, external-secrets operator fetches the client ID (along with other secrets) from the Key Vault and saves it into a k8s secret with a key `AZURE_CLIENT_ID`.

Pods have the following in their spec to populate environment variables:

```yaml
envFrom:
  - secretRef:
      name: <name>
```

aramase:
This is a reasonable change. We can include this in the next release.

PR #1593, opened 2025-04-11
fix(webhook): do not add environment variables when they are empty

weisdd:
**Reason for Change**:

As explained in more details in #1555, results of webhook mutation would be better if empty environment variables are not added to container spec. - `Env` section of a pod spec (used by the webhook) has higher precedence than `envFrom`, so if the environment variables are imported using the latter, container still starts with empty values populated by the webhook.

**Requirements**

- [X] squashed commits
- [ ] included documentation (not applicable)

**Issue Fixed**:
Fixes: #1555

- [ ] yes
- [X] no

**Notes for Reviewers**:
At first, I proposed to make the change only for `AZURE_CLIENT_ID`, @aramase agreed to that in the related issue: https://github.com/Azure/azure-workload-identity/issues/1555#issuecomment-2756202558, but I think it would also make sense to have the same behaviour for `AZURE_TENANT_ID` and `AZURE_AUTHORITY_HOST` even though I personally care only about contents of `AZURE_CLIENT_ID`.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -394,9 +394,9 @@ func addEnvironmentVariables(container corev1.Container, clientID, tenantID, azu
 		{Name: AzureAuthorityHostEnvVar, Value: azureAuthorityHost},
 	}
 
-	// append only the ones not already present
+	// append the ones that are not already present (only if desired env contains a non-empty value)
 	for _, env := range desiredEnvs {
-		if _, ok := m[env.Name]; !ok {
+		if _, ok := m[env.Name]; !ok && env.Value != "" {
 			container.Env = append(container.Env, env)
 		}
 	}
--- a/pkg/webhook/webhook_test.go
+++ b/pkg/webhook/webhook_test.go
@@ -627,6 +627,29 @@ func TestAddEnvironmentVariables(t *testing.T) {
 			}
 		})
 	}
+
+	t.Run("environment variables are not added when empty", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


weisdd:

enj:
I'll defer the review to @aramase but it seems worthwhile to factor out the shared conditional logic for updating the env var slice into a function, especially since we plan on adding more env vars in the future.

aramase:
> I'll defer the review to @aramase but it seems worthwhile to factor out the shared conditional logic for updating the env var slice into a function, especially since we plan on adding more env vars in the future.

This is a valid request! @weisdd are you interested in making this change as part of the PR? If you're busy, we can merge this and I can follow-up with the refactor.

aramase:
> I'll defer the review to @aramase but it seems worthwhile to factor out the shared conditional logic for updating the env var slice into a function, especially since we plan on adding more env vars in the future.

I've refactored the env var injection logic in https://github.com/Azure/azure-workload-identity/pull/1605 to reduce duplication. @weisdd you can rebase on top of that—it should make it easier to apply your non-empty value checks in a single place.

weisdd:
Sure, I'll rebase the PR, thanks :)

enj:
> Sure, I'll rebase the PR once #1605 is merged, thanks :)

The referenced PR is merged now.

weisdd:
@enj @aramase I've rebased the PR, thanks :)

aramase (approved):

Issue #1617, opened 2025-06-06
Support pod anti-affinity, and topologySpreadConstraints in the helm chart


tiesmaster:
**Is your feature request related to a problem? Please describe.**
The problem is that it's not possible to configure pod anti-affinity, and topologySpreadConstraints. This is needed to ensure high-availability for WI so that the pods land on different nodes, or zones. This will allow you to suffer from node, or zone failure.

The problem that we have is that we're running 2 replicas of WI, and both ended up on the same node. We identified this (when researching a different issue, unrelated to WI), and noticed that. It wasn't an issue yet, but we'd like to ensure proper HA for this workload.

**Describe the solution you'd like**
Add support to configure the pod anti-affinity, and topologySpreadConstraints fields in the `Deployment` of the WI pods. This is very common, and should be easy to add. I can open up a PR, if you're interested.

**Describe alternatives you've considered**
There are not really any alternatives here, AFAICT, except for vendoring the helm chart. However, we'd like to support the community, and help improving OSS.

**Additional context**
-

aramase:
close with https://github.com/Azure/azure-workload-identity/pull/1618

PR #1618, opened 2025-06-07
chore(helm chart): Add missing topologySpreadConstraints to the helm chart

tiesmaster:
**Reason for Change**:
The trigger is issue #1617: add `topologySpreadConstraints` in the helm chart

**Requirements**

- [X] squashed commits
- [X] included documentation
- [-] added unit tests and e2e tests (if applicable).

**Issue Fixed**: #1617

- [ ] yes
- [X] no

**Notes for Reviewers**:
Originally, my issue was about adding both pod anti-affinity, and `topologySpreadConstraints`, but I found out that pod anti-affinity can already be done via the `affinity` field. However, as pod anti-affinity cannot handle everything that topologySpreadConstraints can, it still makes sense to add this as option to the helm chart.


aramase (commented):
/azp run

PR #1645, opened 2025-08-11
chore: update to go 1.24.6

aramase:


stlaz:
I suspect azure linux does not have go 1.24.6, hence the failures, right? :disappointed:

PR #1672, opened 2025-09-30
chore: update distroless-iptables to v0.8.2

aramase:


docker/proxy-init.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=${TARGETPLATFORM:-linux/amd64} registry.k8s.io/build-image/distroless-iptables:v0.8.1
```
aramase, line 1: The `--platform` flag is redundant - BuildKit automatically pulls the correct architecture variant from multi-arch base images during `docker buildx build --platform=...` builds. I think with this, we should also see dependabot PRs to automatically bump when newer versions are available.

enj (approved):

PR #1720, opened 2026-04-10
feat: support custom token endpoint injection for workload identity

aramase:
## Summary
  - Add support for the mutating webhook to inject custom token endpoint environment variables (`AZURE_KUBERNETES_TOKEN_PROXY`, `AZURE_KUBERNETES_SNI_NAME`, `AZURE_KUBERNETES_CA_FILE`, `AZURE_KUBERNETES_CA_DATA`) into pods annotated with a configurable annotation.
  - The annotation suffix and audience are configured via webhook CLI flags (`--custom-token-endpoint-annotation-suffix`, `--custom-token-endpoint-audience`). When both are empty (the default), the feature is fully disabled.
  - Support CA distribution via ConfigMap or ClusterTrustBundle projected volume sources.
  - Add Helm chart configuration for all custom token endpoint settings, including the annotation suffix and audience as conditional deployment args.
  - Add e2e test for custom token endpoint webhook mutation.
  - Add `custom-token-endpoint-msal-go` example demonstrating direct token endpoint usage without MSAL.

  ## Breaking Changes

  ### Volume path change

  The projected service account token volume has been renamed and relocated:

  | | Before | After |
  |---|---|---|
  | Volume name | `azure-identity-token` | `azure-workload-identity-reserved-<sha256>` |
  | Mount path | `/var/run/secrets/azure/tokens` | `/var/run/secrets/azure/wi` |
  | Token subpath | `azure-identity-token` | `token/azure-identity-token` |
  | Full token path | `/var/run/secrets/azure/tokens/azure-identity-token` | `/var/run/secrets/azure/wi/token/azure-identity-token` |

  The `AZURE_FEDERATED_TOKEN_FILE` environment variable is updated automatically, so applications using the Azure SDK (which reads this env var) will work without changes. Applications that hardcode the old path will need to be
  updated.


cmd/webhook/main.go
```diff
@@ -80,6 +83,14 @@ func mainErr() error {
 	flag.StringVar(&logLevel, "log-level", "",
 		"In order of increasing verbosity: unset (empty string), info, debug, trace and all.")
 	flag.BoolVar(&versionInfo, "version", false, "Print version information and exit")
+
+	flag.StringVar(&customTokenEndpointAnnotationSuffix, "custom-token-endpoint-annotation-suffix", "identity-binding",
```
enj: I feel like we should not default this.

cmd/webhook/main.go
```diff
@@ -80,6 +83,14 @@ func mainErr() error {
...
+		"Suffix to append to 'azure.workload.identity/use-' when defining a custom token endpoint annotation. "+
+			"Example: 'identity-binding' becomes 'azure.workload.identity/use-identity-binding'.")
+
+	flag.StringVar(&customTokenEndpointAudience, "custom-token-endpoint-audience", "api://AKSIdentityBinding",
```
enj: Same here.

examples/custom-token-endpoint-msal-go/go.mod
enj, line 1: Can we avoid using the word identity binding?

pkg/webhook/webhook.go
```diff
@@ -382,83 +424,120 @@ func getTenantID(sa *corev1.ServiceAccount, c *config.Config) string {
...
 
 	desiredEnvs := []corev1.EnvVar{
 		{Name: AzureClientIDEnvVar, Value: clientID},
 		{Name: AzureTenantIDEnvVar, Value: tenantID},
-		{Name: AzureFederatedTokenFileEnvVar, Value: filepath.Join(TokenFileMountPath, TokenFilePathName)},
+		{Name: AzureFederatedTokenFileEnvVar, Value: filepath.Join(VolumeMountPath, TokenFilePath)},
 		{Name: AzureAuthorityHostEnvVar, Value: azureAuthorityHost},
 	}
 
+	if podUsingIdentityBinding {
```
enj:
```suggestion
	if podUsingCustomTokenEndpoint {
```

enj (commented):
Comments re: IB name usage.

enj (approved):

PR #1721, opened 2026-04-10
refactor: migrate proxy from MSAL to azidentity SDK with LRU credential cache

aramase:
Replace the MSAL confidential client with azidentity.WorkloadIdentityCredential for token acquisition in the proxy sidecar. This simplifies the proxy by:

- Removing manual token file reading (azidentity handles it internally)
- Removing the AZURE_AUTHORITY_HOST requirement (azidentity infers it)
- Removing the readJWTFromFS helper function
- Adding an LRU credential cache (128 entries, 25h TTL) to avoid recreating credentials on every request

The azidentity SDK provides built-in token caching, automatic token file rotation handling, and proper retry logic.


pkg/proxy/cred_cache.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"k8s.io/apimachinery/pkg/util/cache"
+)
+
+const (
+	defaultCacheSize = 128
+	defaultCacheTTL  = 25 * time.Hour // WI token is valid for 24 hours, so set TTL to 25 hours to allow extended usage
```
enj: As in, keep using the credential past its expiration and expect the SDK to handle the 401?

pkg/proxy/proxy.go
```diff
@@ -153,6 +156,28 @@ func (p *proxy) msiHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	wi, err := azidentity.NewWorkloadIdentityCredential(&azidentity.WorkloadIdentityCredentialOptions{
+		ClientID: clientID,
+		TenantID: tenantID,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	p.credCache.Add(key, wi)
```
enj: This should probably single flight to avoid unnecessary `New` calls?

enj (commented):
Minor comments.

pkg/proxy/cred_cache.go
aramase: updated the comment, let me know what you think!

pkg/proxy/proxy.go
aramase: added singleflighting

enj (approved):

PR #1722, opened 2026-04-13
fix: update Go version, remove K8s version table, bump kind versions

aramase:
- Fix `msal-go` Windows image build failure by updating Go base image from 1.25.6 to 1.25.9
- Remove the supported Kubernetes version table from README to reduce maintenance burden
- Update kind node versions in CI pipelines to v1.33.7, v1.34.3, and v1.35.1 (kind v0.31.0)
- Use `kindest/node` directly instead of the `mcr.microsoft.com/mirror` prefix
- Bump `KIND_VER` in Makefile from v0.27.0 to v0.31.0
- Update `wi-kind-setup.sh` default kind image to v1.35.1 and simplify heredoc patterns
- Fix govulncheck CI for `test/e2e` by removing `//go:build e2e` tags (separate module doesn't need them)
- Bump k8s deps in `test/e2e` from v1.31.3 to v1.35.4 to resolve GO-2025-4240


stlaz:
Looks like govulncheck found some actual CVEs in k8s deps.

.github/workflows/scan-vulns.yaml
```diff
@@ -31,10 +31,7 @@ jobs:
       - uses: golang/govulncheck-action@31f7c5463448f83528bd771c2d978d940080c9fd # v1.0.4
 
       - name: Golang vulncheck - test/e2e
-        uses: golang/govulncheck-action@31f7c5463448f83528bd771c2d978d940080c9fd # v1.0.4
-        with:
-          repo-checkout: false
-          work-dir: test/e2e
+        run: govulncheck -C test/e2e -tags e2e -format text ./...
```
stlaz: Can we just remove the build tags from the files in the e2e module?

.pipelines/nightly.yaml
```diff
@@ -81,12 +81,12 @@ jobs:
...
-        kind_v1_31_6:
-          KIND_NODE_VERSION: v1.31.6
-        kind_v1_32_3:
-          KIND_NODE_VERSION: v1.32.3
+        kind_v1_33_7:
+          KIND_NODE_VERSION: v1.33.7
+        kind_v1_34_3:
+          KIND_NODE_VERSION: v1.34.3
+        kind_v1_35_0:
+          KIND_NODE_VERSION: v1.35.0
```
stlaz: 1.35.1 is the newest available kind image now

.pipelines/nightly.yaml
```diff
@@ -81,12 +81,12 @@ jobs:
         value: "true"
     strategy:
       matrix:
```
stlaz, line 83: This matrix definition feels super tedious to maintain, could we switch to something like this?
https://github.com/kubernetes-sigs/secrets-store-sync-controller/blob/2c87a6e44488bb9e256181c4ba84646ad79ac758/.github/workflows/e2e-provider-tests.yaml#L27-L28

scripts/create-kind-cluster.sh
```diff
@@ -57,7 +57,7 @@ EOF
 
 create_kind_cluster() {
   # create a kind cluster
-  cat <<EOF | ${KIND} create cluster --name "${KIND_CLUSTER_NAME}" --image "mcr.microsoft.com/mirror/kindest/node:${KIND_NODE_VERSION:-v1.32.3}" --config=-
+  cat <<EOF | ${KIND} create cluster --name "${KIND_CLUSTER_NAME}" --image "kindest/node:${KIND_NODE_VERSION:-v1.35.0}" --config=-
```
stlaz: why the `cat <<EOF` instead of just
`${KIND} create cluster --name "${KIND_CLUSTER_NAME}" --image "kindest/node:${KIND_NODE_VERSION:-v1.35.0}" --config=- <<EOF`

.pipelines/nightly.yaml
stlaz: Oh, but this is not a GH action but an ADO pipeline and the API is just broken, isn't it?
Never mind then, no action necessary.

scripts/create-kind-cluster.sh
stlaz: There's also kind cluster being created in `scripts/wi-kind-setup.sh` and we may want to bump the kind version in the Makefile to 0.31.0

.github/workflows/scan-vulns.yaml
```diff
@@ -31,10 +31,7 @@ jobs:
       - uses: golang/govulncheck-action@31f7c5463448f83528bd771c2d978d940080c9fd # v1.0.4
 
       - name: Golang vulncheck - test/e2e
-        uses: golang/govulncheck-action@31f7c5463448f83528bd771c2d978d940080c9fd # v1.0.4
-        with:
-          repo-checkout: false
-          work-dir: test/e2e
+        run: govulncheck -C test/e2e -format text ./...
```
stlaz, line 34: The build tag removal should hopefully fix the original GH action failure, making the original config work. Does it?

stlaz: I see in the next commit you made the check optional, that's fine then.

stlaz (approved):

PR #1745, opened 2026-07-17
fix: make projected volume injection idempotent on webhook reinvocation after metadata.name mutation

### LINKED ISSUE #1744: Mutating webhook is not idempotent when reinvoked after another webhook mutates metadata.name, causing duplicate projected volume injection
**Describe the bug**
Users running additional mutating admission webhooks experienced pod admission failures when workloads were deployed through ReplicaSets/Deployments.

The issue appears when the Azure Workload Identity mutating webhook is reinvoked (reinvocationPolicy: IfNeeded) after another mutating webhook modifies pod metadata, causing Workload Identity to inject a second projected volume instead of reusing the previously injected one.

**Steps To Reproduce**
Pods annotated with:
azure.workload.identity/use: "true"
fail admission with an error similar to: "Invalid value: '/var/run/secrets/azure/wi': must be unique"
The pod contains two projected volumes targeting the same mount path: "/var/run/secrets/azure/wi", resulting in Kubernetes rejecting the pod.

1. Install Azure Workload Identity.
Enable webhook reinvocation (IfNeeded).
2. Install a second mutating admission webhook that mutates pod metadata and triggers reinvocation.
3. Create a Deployment using: azure.workload.identity/use: "true"
4. Observe duplicate projected volume injection and pod admission failure.

**Expected behavior**
The webhook should be idempotent across admission reinvocation.

Instead of identifying prior injection through generated names, it should detect existing workload identity projected volumes using an invariant property, for example:
ServiceAccountToken.Path == /var/run/secrets/azure/wi/token
or equivalent workload identity token configuration.

If an existing WI projected volume is already present, the webhook should reuse it rather than injecting another volume.

**Logs**
Example failed deployment logs: kind: Event
apiVersion: v1
metadata:
  name: wi-demo-c68cf6446.18c2dc17a63915c6
  namespace: wi-repro
  uid: 7b6df640-5eb7-4d9f-9637-1ce61171bc3e
  resourceVersion: '9274'
  creationTimestamp: '2026-07-16T19:33:16Z'
  managedFields:
    - manager: kube-controller-manager
      operation: Update
      apiVersion: v1
      time: '2026-07-16T19:33:16Z'
      fieldsType: FieldsV1
      fieldsV1:
        f:count: {}
        f:firstTimestamp: {}
        f:involvedObject: {}
        f:lastTimestamp: {}
        f:message: {}
        f:reason: {}
        f:reportingComponent: {}
        f:source:
          f:component: {}
        f:type: {}
involvedObject:
  kind: ReplicaSet
  namespace: wi-repro
  name: wi-demo-c68cf6446
  uid: 556f2778-68c9-4a81-82c2-a7ba56f75a25
  apiVersion: apps/v1
  resourceVersion: '9264'
reason: FailedCreate
message: >-
  Error creating: Pod "wi-demo-c68cf6446-w449hl" is invalid:
  spec.containers[0].volumeMounts[2].mountPath: Invalid value:
  "/var/run/secrets/azure/wi": must be unique
source:
  component: replicaset-controller
firstTimestamp: '2026-07-16T19:33:16Z'
lastTimestamp: '2026-07-16T19:33:16Z'
count: 1
type: Warning
eventTime: null
reportingComponent: replicaset-controller
reportingInstance: ''
 

**Environment**

- Kubernetes version (use `kubectl version`): K3s
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

**Additional context**

RohanPawarMSFT:
**Reason for Change**:

There is an active issue in AKS: https://github.com/Azure/azure-workload-identity/issues/1744

**Requirements**

- [x] squashed commits
- [ ] included documentation

**Issue Fixed**:

Fixes #1744 

- [ ] yes
- [x] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -231,7 +231,7 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
 	// get containers to skip
 	skipContainers := getSkipContainers(pod)
 	podUsingCustomTokenEndpoint := m.isUsingCustomTokenEndpoint(pod)
-	volumeName := buildVolumeName(podName)
+	volumeName := getOrBuildVolumeName(pod, podName)
 
 	pod.Spec.InitContainers = m.mutateContainers(pod.Spec.InitContainers, clientID, tenantID, skipContainers, podUsingCustomTokenEndpoint, volumeName)
 	pod.Spec.Containers = m.mutateContainers(pod.Spec.Containers, clientID, tenantID, skipContainers, podUsingCustomTokenEndpoint, volumeName)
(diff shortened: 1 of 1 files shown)


enj:
Superseded by #1749

PR #1746, opened 2026-07-20
chore: bump k8s to 1.36 and controller-runtime to v0.24.1

tmshort:
Bumps Kubernetes dependencies from v0.31.14 to v0.36.2 and controller-runtime from v0.19.7 to v0.24.1.

**Dependency changes**

- `k8s.io/api`, `k8s.io/apimachinery`, `k8s.io/client-go`: v0.31.14 → v0.36.2
- `k8s.io/apiextensions-apiserver`, `k8s.io/component-base`: v0.31.0 → v0.36.2
- `sigs.k8s.io/controller-runtime`: v0.19.7 → v0.24.1
- `github.com/open-policy-agent/cert-controller`: v0.11.0 → v0.16.0 (required for controller-runtime v0.24.1 API compatibility)
- `github.com/microsoft/kiota-http-go`: v1.4.7 → v1.5.5 (fixes GO-2026-5224: cookie/auth header leak on cross-host redirect)

**Test fix**

controller-runtime v0.24.1 changed how it handles `creationTimestamp:null` in admission requests: it now emits a spurious `remove /metadata/creationTimestamp` JSON patch when the input contains an explicit `null` for that field. Fixed by stripping `creationTimestamp:null` from test pod JSON in `newPodRaw` and the hardcoded raw pod string, so the input and output representations are consistent and no remove patch is generated. The original strict assertion ("no remove patches") is preserved.

**Verification**

- `go build ./...` passes cleanly
- `go test ./...` all pass


tmshort:
@microsoft-github-policy-service agree company="Red Hat"

tmshort:
@aramase can you re-do `/azp run`?

tmshort:
ping @aramase ?

tmshort:
Sorry, @aramase, I had to bump again due to the various security checks, which I then fix, and then when you do `/azp run`, more seem to be reported. Could you run the rest of the CI again?

tmshort:
ping @aramase for running workflows

stlaz:
/azp run
let me see if I am able to run those

tmshort:
Thank you @stlaz, it looks as though everything is passing!

tmshort:
That last scan failure is due to the golang compiler, and not the code here.
(And it had passed earlier today).

tmshort:
Is there a way to get past this without another update?

tmshort:
ping?

tmshort:
And I have to rebase

tmshort:
Looking for a CI run, and hopefully approvals?

tmshort:
Ping @aramase @enj @stlaz for a CI run and review?

tmshort:
Looking for review?

pkg/webhook/webhook_test.go
```diff
@@ -865,7 +865,7 @@ func TestHandle(t *testing.T) {
 				t.Fatalf("expected to be allowed")
 			}
 			for _, patch := range resp.Patches {
-				if patch.Operation == "remove" {
+				if patch.Operation == "remove" && patch.Path != "/metadata/creationTimestamp" {
```
enj: This is suspicious, we should not be emitting `remove` patches for `creationTimestamp`...

go.mod
```diff
@@ -16,23 +16,30 @@ require (
...
 	monis.app/mlog v0.0.4
-	sigs.k8s.io/controller-runtime v0.19.7
+	sigs.k8s.io/controller-runtime v0.24.1
+)
+
+require (
+	github.com/std-uritemplate/std-uritemplate/go/v2 v2.0.3 // indirect
+	go.yaml.in/yaml/v3 v3.0.4 // indirect
+	sigs.k8s.io/randfill v1.0.0 // indirect
+	sigs.k8s.io/structured-merge-diff/v6 v6.3.2 // indirect
```
enj: Move this with the other `indirect` deps.

stevekuznetsov: Do you `go mod tidy` in CI? Shouldn't that just be automated?

pkg/webhook/webhook_test.go
tmshort: This is due to the new controller-runtime, let me fix an alternative way.

go.mod
tmshort: `go mod tidy` was run locally, and no changes occurred. But it should be cleaned up.

pkg/webhook/webhook_test.go
tmshort: Fixed using an alternative method, but it's due to the new controller-runtime. Without a patch to `webhook_test.go`, the unit tests fail.

tmshort:
@enj I updated based on your comments. Updated the PR and commit description regarding the test fixes required by the controller-runtime updates.

pkg/webhook/webhook_test.go
```diff
@@ -70,6 +71,13 @@ func newPodRaw(name, namespace, serviceAccountName string, labels, annotations m
 	if err != nil {
 		panic(err)
 	}
+	// Strip "creationTimestamp":null from the raw JSON. metav1.Time{} serializes
+	// to null, but when controller-runtime re-encodes the object it omits the
+	// field entirely, generating a spurious "remove /metadata/creationTimestamp"
+	// JSON patch. Removing it here keeps the input and output representations
+	// consistent.
+	raw = bytes.ReplaceAll(raw, []byte(`,"creationTimestamp":null`), nil)
```
timmy-wright: It seems cleaner to parse the JSON and remove the field then serialise again. Doing a byte replacement could cause issues with JSON validity. 

For example, if the "creationTimestamp" happens to be the first field in a JSON blob, this replacement won't work due to the leading comma. But if it's the last you can't check trailing comma. And this will replace creationTimestamp in a string - even if it's not part of the main blob.

tmshort: Correct, that's why there's _two_ `ReplaceAll` lines.

The first one will handle the "all but in the first location" cases, the second one will handle the "first location", since that should be the only option left. So, combined, the two lines  it ought to _all_ cases. However, if it were the _only_ field, then it would fail.

tmshort: But I fixed it to avoid the fragility.

pkg/webhook/webhook_test.go
enj, line 1: @tmshort please undo all changes to this file and update `webhook.go` such that it does not emit these `creationTimestamp` remove patches.

tmshort: @enj done. I was so focused on fixing up the test results, I didn't consider changing the `webhook.go` file.

tmshort:
Looking for another review/CI run

tmshort:
Hi @enj, looking for another review.

tmshort:
Hi @enj, @stlaz, @aramase, looking for another review/CI run?

damdo:
cc. @stevekuznetsov 

Thanks!

aramase:
thanks for the PR, @tmshort!

this was addressed in https://github.com/Azure/azure-workload-identity/pull/1760 and will be part of the upcoming release.

PR #1749, opened 2026-08-03
fix(webhook)!: prevent duplicate token mount path on reinvocation

aramase:
The projected token volume name and mount path derive from the pod identity, which can change between the webhook invocation and its reinvocation (e.g. another mutating webhook populates metadata.name under reinvocationPolicy: IfNeeded). The mount path was a fixed constant, so a second mount was appended at the same path and rejected by the API server with "spec.containers[*].volumeMounts[*].mountPath: must be unique".

Derive the mount path from the same pod identity hash so it is unique per pod and cannot collide across invocations.

BREAKING CHANGE: the projected service account token mount path now includes a per-pod hash segment
(/var/run/secrets/azure/wi/<hash>/token/azure-identity-token). Workloads using the Azure SDK are unaffected since they read AZURE_FEDERATED_TOKEN_FILE; workloads that hardcode the token path must be updated.

**Reason for Change**:

**Requirements**

- [ ] squashed commits
- [ ] included documentation

**Issue Fixed**:

- [ ] yes
- [ ] no

**Notes for Reviewers**:

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -232,9 +232,10 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) (respons
 	skipContainers := getSkipContainers(pod)
 	podUsingCustomTokenEndpoint := m.isUsingCustomTokenEndpoint(pod)
 	volumeName := buildVolumeName(podName)
+	volumeMountPath := buildVolumeMountPath(podName)
 
-	pod.Spec.InitContainers = m.mutateContainers(pod.Spec.InitContainers, clientID, tenantID, skipContainers, podUsingCustomTokenEndpoint, volumeName)
-	pod.Spec.Containers = m.mutateContainers(pod.Spec.Containers, clientID, tenantID, skipContainers, podUsingCustomTokenEndpoint, volumeName)
+	pod.Spec.InitContainers = m.mutateContainers(pod.Spec.InitContainers, clientID, tenantID, skipContainers, podUsingCustomTokenEndpoint, volumeName, volumeMountPath)
+	pod.Spec.Containers = m.mutateContainers(pod.Spec.Containers, clientID, tenantID, skipContainers, podUsingCustomTokenEndpoint, volumeName, volumeMountPath)
 
 	m.addProjectedVolume(pod, serviceAccountTokenExpiration, volumeName, podUsingCustomTokenEndpoint)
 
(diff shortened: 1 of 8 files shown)


pkg/webhook/webhook_test.go
```diff
@@ -596,7 +599,7 @@ func TestAddEnvironmentVariables(t *testing.T) {
 					},
 					{
 						Name:  AzureFederatedTokenFileEnvVar,
-						Value: filepath.Join(VolumeMountPath, TokenFilePath),
+						Value: path.Join(testVolumeMountPath, TokenFilePath),
 					},
```
enj, line 603: nit: would be nicer to assert the raw strings here.

pkg/webhook/webhook.go
```diff
@@ -6,7 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"path/filepath"
+	"path"
```
enj: Why did this change?

pkg/webhook/webhook.go
```diff
@@ -594,8 +595,26 @@ func serverVersionGTE(discoveryClient discovery.ServerVersionInterface, v *utilv
...
+
+// buildVolumeMountPath returns the deterministic mount path for the projected
+// service account token volume. The hash suffix ensures that if the pod name
+// changes between webhook (re)invocations (for example, another mutating webhook
+// populates metadata.name before this webhook is reinvoked under
+// reinvocationPolicy: IfNeeded), the mount lands at a unique path rather than
+// colliding with the previously injected mount, which the API server rejects
+// with "spec.containers[*].volumeMounts[*].mountPath: must be unique".
+func buildVolumeMountPath(podName string) string {
+	return path.Join(VolumeMountPath, volumeHash(podName))
```
enj: Lets rename `VolumeMountPath` to `ProjectedVolumePathPrefix`.

pkg/webhook/webhook.go
```diff
@@ -594,8 +595,26 @@ func serverVersionGTE(discoveryClient discovery.ServerVersionInterface, v *utilv
...
 	hash := sha256.Sum256([]byte(podName))
-	truncatedHash := fmt.Sprintf("%x", hash)[:63-len(ProjectedVolumeNamePrefix)] // 63 is the max length for volume names
-	return fmt.Sprintf("%s%s", ProjectedVolumeNamePrefix, truncatedHash)
+	return fmt.Sprintf("%x", hash)[:63-len(ProjectedVolumeNamePrefix)] // 63 is the max length for volume names
+}
+
+// buildVolumeName returns the deterministic reserved name for the projected
+// service account token volume.
+func buildVolumeName(podName string) string {
+	return fmt.Sprintf("%s%s", ProjectedVolumeNamePrefix, volumeHash(podName))
```
enj:
```suggestion
	return ProjectedVolumeNamePrefix + volumeHash(podName)
```

enj (commented):
Minor comments.

enj (approved):

gil-tohar-cyera:
@enj @aramase Why was this surfaced as a patch level version update (1.6.0 -> 1.6.1) if you are aware it's a breaking change. It should have been included in the change log, and it should have been at least a minor level version update

PR #1757, opened 2026-08-24
fix(webhook): preserve pod-level spec.resources (issue #1751)

### LINKED ISSUE #1751: PodLevelResources not supported (azure-workload-identity doesn't support official AKS versions)
**Describe the bug**
Starting with AKS 1.34 PodLevelResources are available and can be used.
When using with azure-workload-identity the `spec.resources` is removed from the Pod.

It seems that azure-workload-identity is using a very outdated Kubernetes SDK version ([v0.31.14](https://github.com/Azure/azure-workload-identity/blob/main/go.mod#L30)) which doesn't support current AKS GA versions.

Seems that the Pod is serialized into the normal Kubernetes SDK pod struct which doesn't support these features in the MutatingWebhook. Azure-workload-identity should use the dynamic client instead to avoid such issues in future.

**Steps To Reproduce**
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
        azure.workload.identity/use: "true"
    spec:
      # example comes without serviceaccount, a fully setup serviceaccount is still needed for the MutationWebhook
      serviceAccount: workload-identity-serviceaccount 
      containers:
      - name: nginx-container
        image: nginx:latest
      resources:
        limits:
          cpu: 200m
          memory: 200Mi
        requests:
          cpu: 200m
          memory: 200Mi
```
New pods will not have resources on pod level, this breaks eg. Kubernetes `ResourceQuotas`.

**Expected behavior**

Azure-workload-identity should support all AKS GA versions.

**Logs**
no logs needed

**Environment**

- Kubernetes version (use `kubectl version`): v1.35.6
- Cloud provider or hardware configuration: Azure
- OS (e.g: `cat /etc/os-release`): azure-linux

**Additional context**

jiashun0011:
The mutating webhook decoded AdmissionRequests into a typed corev1.Pod struct backed by k8s.io/api v0.31.14, which predates the pod-level spec.resources field (alpha in K8s 1.32, GA in K8s 1.34). As a result, pods that use a workload-identity-enabled ServiceAccount had their pod-level requests/limits silently stripped by the webhook, breaking ResourceQuota and any consumer of pod-level resources.

Bump k8s.io/{api,apimachinery,client-go,apiextensions-apiserver, component-base} and sigs.k8s.io/controller-runtime (and the transitive open-policy-agent/cert-controller) to versions that include the PodLevelResources field, and add a regression test that fails if the webhook ever removes /spec/resources again.

**Reason for Change**:

**Requirements**

- [x] squashed commits
- [ ] included documentation

**Issue Fixed**:

Fixes #1751

- [ ] yes
- [x] no

**Notes for Reviewers**:


enj:
Superseded by #1759

